PR #70, opened 2015-02-09
Adjust where examples exist and remove useless init.pp from manifests directory

cyberious:


joshcooper:
@cyberious why was `manifests/init.pp` removed? I thought every module needed to include an `init.pp` even if it is empty, and @jeffmccune's comments indicate the same.

cyberious:
I conferred with a few people and there is no reason to have it anymore.  In fact I find it confusing for users imho.  If you say ah, `include registry` it does nothing but the user might not understand that and expect it does something.  A little more silly for registry but what about `include sqlserver`, not possible without tell us a little more about what exactly you want.

dreilly1982:
I am confused as to why the ```manifests/init.pp``` was removed, and why ```include registry``` does nothing.  Either way, this probably shouldn't have been a minor version change as it breaks backwards compatibility of code that includes the ```include registry```, ```require registry```, etc.

PR #73, opened 2015-02-24
Overhaul the README

psoloway:
1. Remove the "what this affects" sections.
2. Make sure that for each parameter, where applicable, there is a data
type and a default value.
3. Make sure that for each parameter that's applicable, there is a note
if the parameter is optional.
4. Make sure the links in the README work and are accurate.
5. Update the link in the Contributing section to point here:
https://docs.puppetlabs.com/forge/contributing.html
7. Move all examples to the Usage section.
8. General copyediting.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -4,90 +4,50 @@
 ####Table of Contents
 
 1. [Overview - What is the registry module?](#overview)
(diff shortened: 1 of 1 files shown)


mhaskel:
Dupe of #75

PR #75, opened 2015-03-12
Updates README file per DOC-1492

psoloway:
1. Remove any "what this affects" sections, except where particularly warranted.
2. Make sure that for each parameter, where applicable, there is a data type and a default value.
3. Make sure that for each parameter that's applicable, there is a note if the parameter is optional.
4. Make sure the links in the README work and are accurate.
5. Update the link in the Contributing section to point here: https://docs.puppetlabs.com/forge/contributing.html
6. General copyediting.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -4,62 +4,42 @@
 ####Table of Contents
 
 1. [Overview - What is the registry module?](#overview)
-2. [Setup - The basics of getting started with registry](#setup)
-    * [What registry affects](#what-registry-affects)
-    * [Setup requirements](#setup-requirements)
+2. [Module Description - What registry does and why it is useful](#module-description)
+3. [Setup - The basics of getting started with registry](#setup)
     * [Beginning with registry](#beginning-with-registry)
-        * [Puppet Enterprise console](#puppet-enterprise-console)
-        * [Noteworthy functionality](#noteworthy_functionality)
-3. [Usage - Configuration options and additional functionality](#usage)
-    * [Defined Type: registry::value](#defined-type-registryvalue)
-    * [Custom Type: registry_key](#custom-type-registrykey)
-    * [Custom Type: registry_value](#custom-type-registryvalue)
-    * [Defined Type: registry::service](#defined-type-registryservice)
-    * [Purge Values](#purge-values)
-4. [Limitations - OS compatibility, etc.](#limitations)
-5. [Development - Guide for contributing to the module](#development)
-
+4. [Usage - Configuration options and additional functionality](#usage)
+5. [Reference](#reference)
+    * [Public Defines](#public-defines)
+    * [Public Types](#public-types)
+    * [Parameters](#parameters)
+6. [Limitations](#limitations)
+7. [Development - Guide for contributing to registry](#development)
 
 ##Overview
 
-The registry module enables you to manage your Windows Registry from your *nix puppet master by supplying the types and providers necessary to create and manage Windows Registry keys and values with Puppet. 
+This module supplies the types and providers you'll need to manage the Registry on your Windows nodes.
+
+##Module Description
+
+The Registry is a hierarchical database built into Microsoft Windows. It stores settings and other information for the operating system and a wide range of applications. This module lets Puppet manage individual Registry keys and values, and provides a simplified way to manage Windows services.
 
 ##Setup
 
-###What registry affects
+This module must be installed on your Puppet master. We've tested it with Puppet agents running on Windows Server 2003, 2008 R2, 2012, and 2012 R2.
 
-* files, folders, and services in Windows Registry    
-* node manifests
+###Beginning with registry
 
-###Setup requirements
-
-For this module to run correctly, you must have [pluginsync enabled](http://docs.puppetlabs.com/guides/plugins_in_modules.html#enabling-pluginsync). 
-
-This setting will ensure the types and providers are synchronized and available on the agent before the configuration run takes place. This is the default behavior of the puppet agent on Microsoft Windows platforms. 
-
-###Beginning with Registry
-
-The registry module is intended to enable a *nix master server to interact with a Windows agent. The module must be downloaded and installed on your puppet master.
-
-The bulk of registry's capabilities comes from two resource types: `registry_key` and `registry_value`. Combined, these types allow you to specify the Registry container and the file(s) meant to be in it. 
-
-The defined resource type `registry::value` allows you to manage Registry values and the parent key in one fell swoop.
-
-    registry::value { 'MyApp Setting1':
-      key   => 'HKLM\Software\Vendor\PuppetLabs',
-      value => setting1,
-      data  => 'Hello World!'
-    }
-    
-Within this defined type, you can specify multiple Registry values for one registry key and manage both at once. 
-
-If, on the other hand, you just want to manage a standalone key or value, you can declare each resource type individually in your site.pp manifest.
+Use the `registry_key` type to manage a single registry key:
 
     registry_key { 'HKLM\System\CurrentControlSet\Services\Puppet':
       ensure => present,
     }
 
-and/or
+##Usage
+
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -125,170 +70,157 @@ You can also use this parameter to set a specific registry value as the default
...
-The `registry::service` defined resource type utilizes specific Registry keys and values to manage service entries in the Microsoft service control framework. Specifically, `registry::service` manages the values in the key HKLM\System\CurrentControlSet\Services\$name\.
+#####`ensure`
 
-    registry::service { puppet:
-      ensure       => present,
-      display_name => "Puppet Agent",
-      description  => "Periodically fetches and applies configurations from a Puppet master server.",
-      command      => 'C:\PuppetLabs\Puppet\service\daemon.bat',
-    }
+*Optional.* Tells Puppet whether the key should or shouldn't exist. Valid options: 'present' and 'absent'. Default value: if unspecified, Puppet won't enforce any other parameters unless the key is already present on a node.
```
cyberious: The default is ensure 'present' and will manage the registry keys by default

README.markdown
```diff
@@ -125,170 +70,157 @@ You can also use this parameter to set a specific registry value as the default
...
+*Optional.* Specifies a Registry value for Puppet to manage. Valid options: a string containing a Registry path. If any of the parent keys in the path don't exist, Puppet creates them automatically. For example: 'HKLM\Software' or 'HKEY_LOCAL_MACHINE\Software\Vendor'. Default value: the title of your declared resource.
 
-Specifies the command to execute.
+If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicitly managed using a prefix. For example: '32:HKLM\Software\Value3'.
 
-#####`start`
+#####`ensure`
 
-Specifies the starting mode of the service. Valid values are 'automatic', 'manual', and 'disabled'. 
+*Optional.* Tells Puppet whether the value should or shouldn't exist. Valid options: 'present' and 'absent'. Default value: unspecified (if left unspecified, Puppet won't enforce any other parameters unless the value is already present on a node.)
```
cyberious: Again this is not the case, it is a default of present and will manage the value, default is present

README.markdown
```diff
@@ -125,170 +70,157 @@ You can also use this parameter to set a specific registry value as the default
...
-The Registry module provides a means of ensuring that only Puppet-specified values are associated with a particular key. This is not default behavior, but a feature you must enable.
+####`registry::service`
 
-In order to make sure only the values specified via Puppet are associated
-with a particular key, use the `purge_values => true` parameter of the
-`registry_key` resource. **Enabling this feature will delete any values not managed by Puppet.** 
+#####`ensure`
 
-The `registry::purge_example` class provides a quick and easy way to see a demonstration of how this works. This example class has two modes of operation determined by the Facter fact `PURGE_EXAMPLE_MODE`: 'setup' or 'purge'.
+*Required.* Tells Puppet whether the service should or shouldn't exist. Valid options: 'present' and 'absent'.
```
cyberious: Not required, default is unset which gets changed to 'present' if not specified.

psoloway: Would it be fair to just say "Default value: 'present'"?

README.markdown
```diff
@@ -4,89 +4,50 @@
...
+Use the `registry_key` type to manage a single registry key:
 
     registry_key { 'HKLM\System\CurrentControlSet\Services\Puppet':
       ensure => present,
     }
 
-and/or
+##Usage
+
+The registry module works mainly through two custom resource types: `registry_key` and `registry_value`. These types combine to let you specify a Registry container and its intended contents.
```
joshcooper: The term `custom resource types` gives a negative connotation to me. I get the impression these are hacked together, one-off solutions, not fully supported.

README.markdown
```diff
@@ -4,89 +4,50 @@
...
+
+###Manage a single Registry value
 
     registry_value { 'HKLM\System\CurrentControlSet\Services\Puppet\Description':
       ensure => present,
       type   => string,
-      data   => "The Puppet Agent service periodically manages your configuration",
+      data   => "The Puppet agent service periodically manages your configuration",
```
joshcooper: The name of the service really is "Puppet Agent"

README.markdown
joshcooper: Users shouldn't rely on ensure having a default value. It may be optional for the registry module, but that's not the case in general, and I wouldn't recommend giving users the impression that it is optional. For example, the `file` type doesn't mention its ensure property being optional or not, it just says what the possible values are: https://docs.puppetlabs.com/references/latest/type.html#file-attribute-ensure

psoloway: Updated copy: 

##### `ensure`

Tells Puppet whether the key should or shouldn't exist. Valid options: 'present' and 'absent'. Default value: 'present'.

README.markdown
psoloway: Updated copy: 

##### `ensure`

Tells Puppet whether the value should or shouldn't exist. Valid options: 'present' and 'absent'. Default value: 'present'.

README.markdown
psoloway: Replaced all instances of the phrases "custom resource types" and "custom types" with simply "types". Also replaced "defined types" with "defines".

joshcooper: I think we want to keep "defined types" as that's an actual puppet thing: https://docs.puppetlabs.com/learning/definedtypes.html

psoloway: This represents a global shift we recently decided to make for all new docs. Despite our own reference materials, "define" is already the term most commonly used by the user community. Existing documentation will be updated over time.

README.markdown
cyberious: :+1:

README.markdown
cyberious: Good catch, yeah it is actually reported as "Puppet Agent" and I don't see that changing.

jbondpdx: So, I'm wondering why it is capped. Our style is that we don't cap agent (i.e., "Puppet agent" is correct); is this something we have capitalized somewhere? Or is it a thing Windows is doing?

joshcooper: So I discussed this with Randall (UX) in March 2012, and provided an example:

> C:\>net stop puppet
> The Puppet Agent service is stopping.
> The Puppet Agent service was stopped successfully.

He response was:

> We shouldn't capitalize "Agent."

And I said, 

> all of the display names for other [Windows] services follow this convention:
>
> Automatic Updates
> DHCP Client
> DNS Client
> Distributed Link Tracking Client
> Error Reporting Service
> Indexing Service
> Resultant Set of Policy Provider

and we agreed to leave it as `Puppet Agent`...

jbondpdx: OK, thanks @joshcooper , that's helpful. I don't love it, because it's the only place we would be capitalizing it, which seems weird. I'm going to knock the question around a little more in Docs, but it sounds like we could be sort of stuck.

laurenrother: @joshcooper I'm curious whether it is in a message as Puppet Agent or whether there is a service called PuppetAgent (or whatever) in the code, or perhaps some variation of Puppet Agent as the name of something in the code itself, outside of a message.

joshcooper: @laurenrother @jbondpdx Windows services have both a short name and a display name. The short name is often used in code, e.g. when programmatically controlling the service `net start puppet`. The short name is usually all lowercase, is US-ASCII, and for us is `puppet` (or `pe-puppet`).

The display name is displayed in Windows GUIs, is often localized, and is usually uppercased as a title, e.g. `Puppet Agent`. The display name also appears in messages, e.g. from above `The Puppet Agent service is stopping.` Note the message is emitted by the service control manager (and not us) using the service's display name.

laurenrother: Ah. Well, if it is in the Windows GUI that way and all of that is controlled by Microsoft, we're out of luck. I discussed this with @jbondpdx and @MFredette , and we agree that given the way it is presented to the user via GUI, etc. let's change the capitalization to Puppet Agent, @psoloway 

Thank you for all your time, patience, and feedback, @joshcooper !

psoloway: Roger wilco.

PR #80, opened 2015-05-08
(MODULES-422) HKEY_USERS support

solcates:
Added basic HKEY_USERS support, and a custom fact $windows_currentuser_sid to be used in conjunction with HKEY_USERS to simulate a HKEY_CURRENT_USER operation.

It's a bit of a hack, but i could see adding other functions such as user and domain parameters to look up the SID on a given "hku" resource.

We are using it for some basic HKCU settings for Office 2013 settings, and seems to work great both for keys and values.

Cheers,

Sol


lib/puppet_x/puppetlabs/registry/provider_base.rb
```diff
@@ -128,6 +128,8 @@ def hkeys
         {
           :hkcr => Win32::Registry::HKEY_CLASSES_ROOT,
           :hklm => Win32::Registry::HKEY_LOCAL_MACHINE,
+          :hku  => Win32::Registry::HKEY_USERS,
+          :hkcu => Win32::Registry::HKEY_USERS,
```
ferventcoder: Is this a typo?

solcates: Bah, busted.  I left that by accident when i was tinkering with the idea of a "hcku" root, but ended up just doing "hku" + a custom fact to get the current user, or supply your own.  I'll fix

solcates: fixed, committed, pushed.

joshcooper:
Hey @solcates, small world~ Thanks for your PR. I had a few comment inline.

examples/users_exmaple.pp
```diff
@@ -0,0 +1,8 @@
+registry_key { "HKU\\${windows_currentuser_sid}\\Test2":
+  ensure => present
+}->
```
joshcooper: This example shows the user as part of the key path, but below you've added the `user` as a new parameter to the registry_key.

solcates: @joshcooper Small world, indeed.  Just don't laugh at my ruby code, i'm still getting over my python bias :)

yeah, i started to go down the path of having the module do a SID lookup locally via WMI, but pulled back, and thus i should have removed the param.  

I'm happy to try and get the SID supplied via params (username, domain), but I know best practice is to usually use facts where possible. Advice ?

lib/facter/currentuser.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+    if username.nil?
+      sid = nil
+    else
+      query2 = "select sid from win32_useraccount where name='"+ name +"' and domain='"+ domain +"'"
+      wmi.ExecQuery(query2).each do |data|
+        sid = data.SID
+      end
+    end
+    username
```
joshcooper: Facter has a core `id` fact that should return the `"domain\\user"` style name, or the short name for built-in accounts. You could just update the registry provider to do the conversion to sid by calling https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/sid.rb#L16

One thing we've been bitten before is using WMI to query account names and not restricting the query to local accounts.

solcates: @joshcooper thanks for the hint on `id`, however it will pull in the id of the process that is running puppet, not the current console user (which is handy for "HKCU" support).

I'l look at using the sid converter instead of wmi, but may still rely on win32_computersystem to supply the "username" of the current console user.

Also i'll add support for 2 parameters `user` and `domain`, and use the sid conversion instead of WMI.  should simplify things for HKU use cases.

Thanks for the advice bud.

ferventcoder:
Where is this one at currently?

solcates:
@ferventcoder buried under the pile of other work unfortunately.

I'm hoping to have some cycles in a week or so to clean this up.

ferventcoder:
@solcates thanks for the update!

tphoney:
How is the work going ? @solcates

tphoney:
@ferventcoder Are these changes ok with you ?
@solcates these commits could do with a squash.

ferventcoder:
@solcates Please don't use merge. It creates an unnecessary merge commit. We recommend the following work stream if you need to update your changes (the key here is ***if***, many times it is not necessary):

`git checkout master`
`git fetch puppetlabs`
`git rebase puppetlabs/master`
`git checkout hkey_users_support`
`git rebase master`
`git push origin hkey_users_support -f`

Where *puppetlabs* is the name of the remote to this repo and *origin* is the name of your remote.

I did check https://docs.puppetlabs.com/forge/contributing.html#making-changes and we don't explicitly state it here, I did followup with a ticket to have it added.

jonnytpuppet:
Hi There

I'm closing this PR due to inactivity. Please feel free to re-open it if you are still interested in this feature.

Thanks!

dandunckelman:
@solcates please reopen.  We're having to work around this, which isn't ideal.

Thanks

ferventcoder:
@dandunckelman if you are able to provide a fix, we would love to evaluate it. 

Here's a module that would give you a great example of how to actually implement this - https://github.com/badgerious/puppet-windows-env/blob/master/lib/puppet/provider/windows_env/windows_env.rb

ferventcoder:
Clarification - on how it could load and unload user registry hives.

PR #85, opened 2015-06-30
(maint) Relax gem dependencies for Windows

ferventcoder:
The version of Puppet/Facter/Hiera will determine the exact versions
of the Windows gems, so relax the versioning a bit to allow for the
proper versions of gems to be pulled in.

There is no need to get super specific on x86 versus x64, the Puppet
gem will determine the exact version (based on that we were shipping
x64 versions of Puppet with the proper dependencies, so we're not
subject to the issues that were present when we didn't have Windows
specific Puppet gems).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,12 +1,27 @@
 source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
-def location_for(place, fake_version = nil)
-  if place =~ /^(git[:@][^#]*)#(.*)/
+# Determines what type of gem is requested based on place_or_version.
+def gem_type(place_or_version)
+  if place_or_version =~ /^git:/
+    :git
+  elsif place_or_version =~ /^file:/
+    :file
+  else
+    :gem
+  end
+end
+
+# Find a location or specific version for a gem. place_or_version can be a
+# version, which is most often used. It can also be git, which is specified as
+# `git://somewhere.git#branch`. You can also use a file source location, which
+# is specified as `file://some/location/on/disk`.
+def location_for(place_or_version, fake_version = nil)
+  if place_or_version =~ /^(git[:@][^#]*)#(.*)/
     [fake_version, { :git => $1, :branch => $2, :require => false }].compact
-  elsif place =~ /^file:\/\/(.*)/
+  elsif place_or_version =~ /^file:\/\/(.*)/
     ['>= 0', { :path => File.expand_path($1), :require => false }]
   else
-    [place, { :require => false }]
+    [place_or_version, { :require => false }]
   end
 end
 
@@ -28,31 +43,62 @@ group :system_tests do
   gem 'beaker-puppet_install_helper', :require => false
 end
 
-is_x64 = Gem::Platform.local.cpu == 'x64'
-if is_x64
-  platform(:x64_mingw) do
-    gem "win32-dir", "~> 0.4.9", :require => false
-    gem "win32-process", "~> 0.7.4", :require => false
-    gem "win32-service", "~> 0.8.4", :require => false
-    gem "minitar", "~> 0.5.4", :require => false
-  end
-else
-  platform(:mingw) do
-    gem "win32-process", "~> 0.6.5", :require => false
-    gem "win32-service", "~> 0.7.2", :require => false
-    gem "minitar", "~> 0.5.4", :require => false
-    if RUBY_VERSION =~ /^1\./ then
-      gem "win32console", :require => false
+# The recommendation is for PROJECT_GEM_VERSION, although there are older ways
+# of referencing these. Add them all for compatibility reasons. We'll remove
+# later when no issues are known. We'll prefer them in the right order.
+puppetversion = ENV['PUPPET_GEM_VERSION'] || ENV['GEM_PUPPET_VERSION'] || ENV['PUPPET_LOCATION'] || '>= 0'
+gem 'puppet', *location_for(puppetversion)
+
+# Only explicitly specify Facter/Hiera if a version has been specified.
+# Otherwise it can lead to strange bundler behavior. If you are seeing this
+# behavior, set `DEBUG_RESOLVER` environment variable to `1` and run bundle
+# install.
+facterversion = ENV['FACTER_GEM_VERSION'] || ENV['GEM_FACTER_VERSION'] || ENV['FACTER_LOCATION']
+gem "facter", *location_for(facterversion) if facterversion
+hieraversion = ENV['HIERA_GEM_VERSION'] || ENV['GEM_HIERA_VERSION'] || ENV['HIERA_LOCATION']
+gem "hiera", *location_for(hieraversion) if hieraversion
+
+
+# For Windows dependencies, these could be required based on the version of
+# Puppet you are requiring. Anything greater than v3.5.0 is going to have
+# Windows-specific dependencies dictated by the gem itself. The other scenario
+# is when you are faking out Puppet to use a local file path / git path.
+explicitly_require_windows_gems = false
+puppet_gem_location = gem_type(puppetversion)
+# This is not a perfect answer to the version check
+if puppet_gem_location != :gem || puppetversion < '3.5.0'
+  is_x64 = Gem::Platform.local.cpu == 'x64'
+  if is_x64
+    platform(:x64_mingw) do
+      explicitly_require_windows_gems = true
+    end
+  else
+    platform (:mingw) do
+      explicitly_require_windows_gems = true
     end
   end
 end
 
-ENV['GEM_PUPPET_VERSION'] ||= ENV['PUPPET_GEM_VERSION']
-puppetversion = ENV['GEM_PUPPET_VERSION']
-if puppetversion
-  gem 'puppet', *location_for(puppetversion)
-else
-  gem 'puppet', :require => false
+if explicitly_require_windows_gems
+  gem "ffi", "~> 1.9.0", :require => false
+  gem "win32-dir", "~> 0.3", :require => false
+  gem "win32-eventlog", "~> 0.5", :require => false
+  gem "win32-process", "~> 0.6", :require => false
+  gem "win32-security", "~> 0.1", :require => false
+  gem "win32-service", "~> 0.7", :require => false
+  gem "minitar", "~> 0.5.4", :require => false
+
+  if RUBY_VERSION =~ /^1\./ then
+    gem "win32console", :require => false
+  end
+
+  # Puppet less than 3.7.0 requires these.
+  # Puppet 3.5.0+ will control the actual requirements.
+  gem "sys-admin", "~> 1.5", :require => false
+  gem "win32-api", "~> 1.4.8", :require => false
+  gem "win32-taskscheduler", "~> 0.2", :require => false
+  gem "windows-api", "~> 0.4", :require => false
+  gem "windows-pr", "~> 1.2", :require => false
 end
 
 if File.exists? "#{__FILE__}.local"


Gemfile
```diff
@@ -28,22 +28,14 @@ group :system_tests do
...
-  end
-else
-  platform(:mingw) do
-    gem "win32-process", "~> 0.6.5", :require => false
-    gem "win32-service", "~> 0.7.2", :require => false
-    gem "minitar", "~> 0.5.4", :require => false
-    if RUBY_VERSION =~ /^1\./ then
-      gem "win32console", :require => false
-    end
+platforms :mswin, :mingw do
```
joshcooper: If you're running on 64-bit ruby, I don't know that it will match either of these. Should that be `:x64_mingw, :mingw`? And that reminds me that `:x64_mingw` may not be a recognized identifier on 32-bit ruby, and bundler will barf when it tries to install... So you may want to keep the `Gem::Platform.local.cpu` check.

Iristyle: Yup, @joshcooper is right... rewriting this is actually dangerous based on older Bundler versions.

ferventcoder: >If you're running on 64-bit ruby, I don't know that it will match either of these.

From the commit message "There is no need to get super specific on x86 versus x64, the Puppet
gem will determine the exact version"

In general the modules don't need to be as specific and can't be if you go across enough versions of Puppet. In all honesty, almost none of this is required anymore once you go above a certain version of Puppet. Once the earliest version we support is Puppet 3.5.x+, we can remove this section entirely.

> rewriting this is actually dangerous based on older Bundler versions.

@Iristyle older Bundler versions won't recognize the `x64_mingw` so I'm not sure I understand. I didn't put back in x64_mingw, so I don't see how what is up for review is dangerous.

ferventcoder: In fact, nearly this exact configuration is up with puppetlabs-acl - https://github.com/puppetlabs/puppetlabs-acl/blob/master/Gemfile#L40-L54

If the above was dangerous or didn't work, well then I'm not sure that AppVeyor agrees as it is running on 3.4.3, 3.7.5 x86/x64 (needs updated to 3.8.1), and 4.0.0 x86/x64 (needs updated to 4.2.0)  https://github.com/puppetlabs/puppetlabs-acl/blob/master/appveyor.yml#L13-L22

https://ci.appveyor.com/project/puppetlabs/puppetlabs-acl (take a look at each matrix, it outputs Environment vars, bundler, and gemfile.lock).

ferventcoder: @joshcooper 

Puppet controls the actual dependencies once you are above 3.5.x, not this gemfile.
```
    puppet (4.2.0-x64-mingw32)
      facter (> 2.0, < 4)
      ffi (~> 1.9.6)
      hiera (>= 2.0, < 3)
      json_pure
      minitar (~> 0.5.4)
      win32-dir (~> 0.4.9)
      win32-eventlog (~> 0.6.2)
      win32-process (~> 0.7.4)
      win32-security (~> 0.2.5)
      win32-service (~> 0.8.6)
```

ferventcoder: Now, going back to 3.3.0, because that is the earliest this module supports:

```
c:\code\puppetlabs-registry> SET GEM_PUPPET_VERSION=3.3.0
c:\code\puppetlabs-registry> ruby -v
ruby 1.9.3p545 (2014-02-24) [i386-mingw32]
c:\code\puppetlabs-registry> bundle install
c:\code\puppetlabs-registry> type Gemfile.lock
    puppet (3.3.0)
      facter (~> 1.6)
      hiera (~> 1.0)
      rgen (~> 0.6.5)
    win32-dir (0.5.0)
      ffi (>= 1.0.0)
    win32-process (0.8.0)
      ffi (>= 1.0.0)
    win32-security (0.2.5)
      ffi
    win32-service (0.8.6)
      ffi
    win32console (1.3.2-x86-mingw32)

PLATFORMS
  x86-mingw32
```

Note that the Gemfile is now doing more control of the dependencies.

Also note that the versions of the Gems are not technically correct. However to become more correct, we'd need to add the Gemfile dependencies in a nice complicated scenario based on each Puppet version below 3.5.x and I just would rather let them move up in this way and deal with errors if we come across any in specs (which haven't come up yet as valid reasons for failure in any place I've converted). Since we don't test against the oldest version supported, this section is really for any community members who may be.

ferventcoder: This does however point out that I need to add more because in 3.3.0 we had more things to depend on.

ferventcoder: Those items have been added. This now works correctly against all versions of Puppet that are supported by this module.

ferventcoder:
@justinstoller I found something else that could be breaking specs. It appears someone left a pry in the code. :/

Will add that to this PR.

ferventcoder:
The tl;dr - We don't need to worry about `:x64_mingw` for Puppet 3.6.x and below (because only x86), and Puppet 3.5.x and above will pull in the right dependencies, mostly ignoring the windows-specific section - unless the windows section is too restrictive, like it is before this PR. Then it errors. This makes the errors go away and everyone is happy.

ferventcoder:
Perhaps the constraint on the section should check the Puppet version to make it more clear?

Gemfile
Iristyle: Sorry @ferventcoder - I left a drive-by "me too" comment on here as I was rushing out the door, without offering up a lot of explanation.  I apologize if that seemed offensive, and I apologize for not being clear about the "dangerous" part (which would be using the `:x64_mingw` symbol without the current guard in place).  For some more background on that goofy guard, see:

https://github.com/puppetlabs/puppet/commit/f6eb2afe292902d1fccaeadb71425da9ee6d12be
https://github.com/puppetlabs/puppet/commit/b6a20930d589daadc36f7681fc40b1a0c6bb29c3

The problem is that `:mswin` doesn't actually match our `x64` Ruby runtime, and won't install platform specific gems in an `x64-mingw32` environment.  See [bundler code](https://github.com/bundler/bundler/blob/f4598e4a01d66677dbd54c9b1d8fd6784daf7271/lib/bundler/current_ruby.rb) for how the platforms are determined - it can be surprising.  (As you've noted, you're seeing the desired end result, but it's a bit of a coincide that it works - the `:mswin` part is meaningless)

You can simulate this with a quick test inside and x64 Ruby with this `Gemfile`.

```ruby
source "https://rubygems.org"

gem "ffi", "1.9.8", :platforms => [:mswin]
```

```
C:\source\x64-gem-test> ruby --version
ruby 2.1.5p273 (2014-11-13 revision 48405) [x64-mingw32]

C:\source\x64-gem-test> bundle install --path .bundle\gems
DL is deprecated, please use Fiddle
Fetching gem metadata from https://rubygems.org/...........
Resolving dependencies...
Using bundler 1.7.12
Your bundle is complete!
It was installed into ./.bundle/gems

C:\source\x64-gem-test> type .\Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:

PLATFORMS
  x64-mingw32

DEPENDENCIES
  ffi (= 1.9.8)
```

So as you can see, the `ffi` gem doesn't actually get installed, because the Bundler platform is not a match.


To illustrate the correct behavior, you could change the `Gemfile`

```ruby
source "https://rubygems.org"

gem "ffi", "1.9.8", :platforms => [:x64_mingw]
```

Clear the lock, and .. voila, `ffi` is installed.

```
C:\source\x64-gem-test> del Gemfile.lock
C:\source\x64-gem-test> del .\.bundle\ -Recurse -Force
C:\source\x64-gem-test> bundle install --path .bundle\gems
DL is deprecated, please use Fiddle
Fetching gem metadata from https://rubygems.org/...........
Resolving dependencies...
Using bundler 1.7.12
Installing ffi 1.9.8
Your bundle is complete!
It was installed into ./.bundle/gems

C:\source\x64-gem-test> type Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    ffi (1.9.8-x64-mingw32)

PLATFORMS
  x64-mingw32

DEPENDENCIES
  ffi (= 1.9.8)
```

I agree that gem dependencies should be picked up automatically by the inclusion of `Puppet` itself (which means that we shouldn't have to include any win32 dependencies).  I think that you've discovered one of the reasons that this exists in the first place is Puppet 3.3.0.  I think another reason is that we didn't start shipping platform gems until `3.5.0` - see https://rubygems.org/gems/puppet/versions

I'm wondering if there was another historical reason that the `Gemfile` looks like this that I'm forgetting.  It might make sense to guard all of this nonsense based on Puppet gem version if that's the rationale.  I think being explicit there could make it much clearer to future maintainers.

ferventcoder: > The problem is that :mswin doesn't actually match our x64 Ruby runtime, and won't install platform specific gems in an x64-mingw32 environment. See bundler code for how the platforms are determined - it can be surprising. (As you've noted, you're seeing the desired end result, but it's a bit of a coincide that it works - the :mswin part is meaningless)

Not really a coincidence based on what I laid out in my explanation. But as I mentioned, making it more obvious may be what is necessary to show why it is not a coincidence.

joshcooper: > I'm wondering if there was another historical reason that the Gemfile looks like this that I'm forgetting

IIRC it looked that way because if you referenced `:x64_mingw` in the Gemfile, then older versions of bundler (1.3.5) running on 32-bit ruby would cause `bundle install` to explode, since it didn't recognize that platform identifier. We worked around this in a few different commits, but this one looks most relevant: https://github.com/puppetlabs/puppet/commit/b6a20930d589daadc36f7681fc40b1a0c6bb29c3.

That said it may not be required anymore, assuming we're using a newer version of bundler everywhere.

Iristyle: Just as a quick follow up on this... 

I think we were able to determine that these gems are also listed here so that you can set `ENV['GEM_PUPPET_VERSION']` (see https://github.com/ferventcoder/puppetlabs-registry/blob/maint/master/gemfile-windows-relax-compatibility/Gemfile#L50-L56) to `file:///` so that you might test local unpublished Puppet changes in concert with changes to a gem.

Puppets `.gemspec` doesn't include any of the platform specific gems in https://github.com/puppetlabs/puppet/blob/master/.gemspec ... so without them in the `Gemfile`, there will be missing gems.

Supporting this workflow was key during some of the FFI development work IIRC.

Using 484cf18 (which is not the latest commit pushed anymore) and pointing to locally cloned Puppet source.

```
C:\source\puppetlabs-registry [pr/85]> ruby --version
ruby 2.1.5p273 (2014-11-13 revision 48405) [x64-mingw32]

C:\source\puppetlabs-registry [pr/85]> $ENV:GEM_PUPPET_VERSION = 'file://c:\source\puppet'

C:\source\puppetlabs-registry [pr/85]> bundle install --path .bundle\gems
DL is deprecated, please use Fiddle
Fetching gem metadata from https://rubygems.org/..........
Resolving dependencies...
Installing CFPropertyList 2.3.1
Installing i18n 0.7.0
Installing rake 10.4.2
Installing minitest 5.7.0
Installing thread_safe 0.3.5
Installing addressable 2.3.8
Installing multi_json 1.11.1
Installing extlib 0.9.16
Installing mini_portile 0.6.2
Installing builder 3.2.2
Installing excon 0.45.3
Installing formatador 0.2.5
Installing json 1.8.3
Installing mime-types 2.6.1
Installing net-ssh 2.9.2
Installing ipaddress 0.8.0
Installing inflecto 0.0.2
Installing retriable 1.4.1
Installing multipart-post 2.0.0
Installing jwt 1.5.1
Installing open_uri_redirections 0.2.1
Installing hocon 0.9.2
Installing inifile 2.0.2
Installing trollop 2.1.2
Installing diff-lcs 1.2.5
Installing unf_ext 0.0.7.1
Installing json_pure 1.8.2
Installing metaclass 0.0.4
Installing rsync 1.0.9
Installing puppet-lint 1.1.0
Installing rspec-mocks 2.14.6
Using bundler 1.7.12
Installing rspec-core 2.14.8
Installing fission 0.5.0
Installing puppet-syntax 2.0.0
Installing launchy 2.4.3
Installing tzinfo 1.2.2
Installing autoparse 0.3.3
Installing docker-api 1.21.4
Installing nokogiri 1.6.6.2
Installing net-scp 1.2.1
Installing faraday 0.9.1
Installing rspec-expectations 2.14.5
Installing unf 0.1.4
Installing mocha 0.10.5
Installing activesupport 4.2.3
Installing rbvmomi 1.8.2
Installing fog-core 1.31.1
Installing aws-sdk-v1 1.64.0
Installing signet 0.6.1
Installing rspec 2.14.1
Installing fog-json 1.0.2
Installing fog-xml 0.1.2
Installing fog-local 0.2.1
Installing fog-vmfusion 0.1.0
Installing aws-sdk 1.64.0
Installing rspec-puppet 2.2.0
Installing google-api-client 0.8.2
Installing fog-sakuracloud 1.0.1
Installing fog-brightbox 0.7.2
Installing fog-serverlove 0.1.2
Installing fog-softlayer 0.4.7
Installing fog-atmos 0.1.0
Installing fog-storm_on_demand 0.1.1
Installing fog-ecloud 0.3.0
Installing fog-google 0.0.6
Installing fog-aws 0.6.0
Installing fog-powerdns 0.1.1
Installing fog-profitbricks 0.0.3
Installing fog-riakcs 0.1.0
Installing fog-radosgw 0.0.4
Installing fog-terremark 0.1.0
Installing fog-voxel 0.1.0
Installing puppetlabs_spec_helper 0.10.3
Installing fog 1.31.0
Installing beaker 2.15.1
Installing beaker-puppet_install_helper 0.2.1
Installing ffi 1.9.10
Installing win32-dir 0.4.9
Installing win32-security 0.2.5
Installing hiera 2.0.0
Installing facter 2.4.4
Using puppet 4.1.0 from source at c:/source/puppet
Your bundle is complete!
It was installed into ./.bundle/gems
Post-install message from nokogiri:
Nokogiri is built with the packaged libraries: libxml2-2.9.2, libxslt-1.1.28, zlib-1.2.8, libiconv-1.14.
```

This shows that `win32-dir` and `win32-security` get picked up (through `Facter 2.4.4` as a transitive dependency), but `win32-process` does not... (because `:mswin` is used as the gem platform)

Dependency management, don't you :heart: it?

Gemfile
```diff
@@ -28,31 +28,62 @@ group :system_tests do
...
+# Puppet Gem to use a local file path / git path
+explicitly_require_windows_gems = false
+if ENV['PUPPET_LOCATION'] || (ENV['GEM_PUPPET_VERSION'] && ENV['GEM_PUPPET_VERSION'] < '3.5.0')
+  is_x64 = Gem::Platform.local.cpu == 'x64'
+  if is_x64
+    platform(:x64_mingw) do
+      explicitly_require_windows_gems = true
+    end
+  else
+    platforms :mswin, :mingw do
```
Iristyle: I don't think that we have any `:mswin` platforms, do we?

Gemfile
```diff
@@ -28,31 +28,62 @@ group :system_tests do
...
-puppetversion = ENV['GEM_PUPPET_VERSION']
-if puppetversion
-  gem 'puppet', *location_for(puppetversion)
-else
-  gem 'puppet', :require => false
+if explicitly_require_windows_gems
+  gem "ffi", "~> 1.9.0", :require => false
+  gem "win32-dir", "~> 0.3", :require => false
+  gem "win32-eventlog", "~> 0.5", :require => false
+  gem "win32-process", "~> 0.6", :require => false
```
Iristyle: So I think that something isn't quite right here when verifying with Puppet from source (on dc39fe2)

```
C:\source\puppetlabs-registry [pr/85]> ruby --version
ruby 2.1.5p273 (2014-11-13 revision 48405) [x64-mingw32]

C:\source\puppetlabs-registry [pr/85]> echo $env:GEM_PUPPET_VERSION
file://c:\source\puppet

C:\source\puppetlabs-registry [pr/85]> del gemfile.lock
C:\source\puppetlabs-registry [pr/85]> del .bundle -Recurse -Force

C:\source\puppetlabs-registry [pr/85]> git lastref
dc39fe2

C:\source\puppetlabs-registry [pr/85]> bundle install --path .bundle\gems
DL is deprecated, please use Fiddle
Fetching gem metadata from https://rubygems.org/..........
Resolving dependencies...
Installing CFPropertyList 2.3.1
Installing i18n 0.7.0
Installing rake 10.4.2
Installing minitest 5.7.0
Installing thread_safe 0.3.5
Installing addressable 2.3.8
Installing extlib 0.9.16
Installing multi_json 1.11.1
Installing mini_portile 0.6.2
Installing excon 0.45.3
Installing builder 3.2.2
Installing mime-types 2.6.1
Installing formatador 0.2.5
Installing ipaddress 0.8.0
Installing inflecto 0.0.2
Installing net-ssh 2.9.2
Installing multipart-post 2.0.0
Installing retriable 1.4.1
Installing jwt 1.5.1
Installing json 1.8.3
Installing inifile 2.0.2
Installing hocon 0.9.2
Installing open_uri_redirections 0.2.1
Installing trollop 2.1.2
Installing rsync 1.0.9
Installing diff-lcs 1.2.5
Installing unf_ext 0.0.7.1
Installing metaclass 0.0.4
Installing json_pure 1.8.2
Installing puppet-lint 1.1.0
Installing rspec-mocks 2.14.6
Using bundler 1.7.12
Installing rspec-core 2.14.8
Installing puppet-syntax 2.0.0
Installing tzinfo 1.2.2
Installing fission 0.5.0
Installing launchy 2.4.3
Installing autoparse 0.3.3
Installing net-scp 1.2.1
Installing nokogiri 1.6.6.2
Installing faraday 0.9.1
Installing rspec-expectations 2.14.5
Installing docker-api 1.21.4
Installing unf 0.1.4
Installing mocha 0.10.5
Installing fog-core 1.31.1
Installing activesupport 4.2.3
Installing rbvmomi 1.8.2
Installing aws-sdk-v1 1.64.0
Installing rspec 2.14.1
Installing signet 0.6.1
Installing fog-xml 0.1.2
Installing fog-json 1.0.2
Installing fog-local 0.2.1
Installing fog-vmfusion 0.1.0
Installing aws-sdk 1.64.0
Installing rspec-puppet 2.2.0
Installing google-api-client 0.8.2
Installing fog-atmos 0.1.0
Installing fog-profitbricks 0.0.3
Installing fog-terremark 0.1.0
Installing fog-ecloud 0.3.0
Installing fog-voxel 0.1.0
Installing fog-brightbox 0.7.2
Installing fog-google 0.0.6
Installing fog-powerdns 0.1.1
Installing fog-radosgw 0.0.4
Installing fog-aws 0.6.0
Installing fog-riakcs 0.1.0
Installing fog-sakuracloud 1.0.1
Installing fog-serverlove 0.1.2
Installing fog-softlayer 0.4.7
Installing fog-storm_on_demand 0.1.1
Installing puppetlabs_spec_helper 0.10.3
Installing fog 1.31.0
Installing beaker 2.15.1
Installing beaker-puppet_install_helper 0.2.1
Installing ffi 1.9.10
Installing win32-dir 0.4.9
Installing win32-security 0.2.5
Installing hiera 2.0.0
Installing facter 2.4.4
Using puppet 4.1.0 from source at c:/source/puppet
Your bundle is complete!
It was installed into ./.bundle/gems
Post-install message from nokogiri:
Nokogiri is built with the packaged libraries: libxml2-2.9.2, libxslt-1.1.28, zlib-1.2.8, libiconv-1.14.
```

A quick test shows that `win32-process` hasn't installed:

```
C:\source\puppetlabs-registry [pr/85]> bundle exec ruby -e "require 'puppet'"
DL is deprecated, please use Fiddle
c:/source/puppet/lib/puppet/util/windows/process.rb:2:in `require': cannot load such file -- win32/process (LoadError)
        from c:/source/puppet/lib/puppet/util/windows/process.rb:2:in `<top (required)>'
        from c:/source/puppet/lib/puppet/util/windows.rb:20:in `require'
        from c:/source/puppet/lib/puppet/util/windows.rb:20:in `<module:Windows>'
        from c:/source/puppet/lib/puppet/util/windows.rb:1:in `<top (required)>'
        from c:/source/puppet/lib/puppet/file_system/windows.rb:2:in `require'
        from c:/source/puppet/lib/puppet/file_system/windows.rb:2:in `<top (required)>'
        from c:/source/puppet/lib/puppet/file_system.rb:10:in `require'
        from c:/source/puppet/lib/puppet/file_system.rb:10:in `<module:FileSystem>'
        from c:/source/puppet/lib/puppet/file_system.rb:1:in `<top (required)>'
        from c:/source/puppet/lib/puppet/file_system/uniquefile.rb:1:in `require'
        from c:/source/puppet/lib/puppet/file_system/uniquefile.rb:1:in `<top (required)>'
        from c:/source/puppet/lib/puppet/util.rb:11:in `require'
        from c:/source/puppet/lib/puppet/util.rb:11:in `<top (required)>'
        from c:/source/puppet/lib/puppet.rb:12:in `require'
        from c:/source/puppet/lib/puppet.rb:12:in `<top (required)>'
        from -e:1:in `require'
        from -e:1:in `<main>'
```

Gemfile
```diff
@@ -28,31 +28,62 @@ group :system_tests do
...
+    gem "win32-taskscheduler", "~> 0.2", :require => false
+    gem "windows-api", "~> 0.4", :require => false
+    gem "windows-pr", "~> 1.2", :require => false
+  end
+
+  # If you are using file locations for Puppet it
+  # won't catch these two dependencies, so we will
+  # want those as well
+  gem "facter", *location_for(ENV['FACTER_LOCATION'] || '>= 1')
+  gem "hiera", *location_for(ENV['HIERA_LOCATION'] || '>= 1')
```
Iristyle: Hmm, in my experience with using Puppet locally from disk via `GEM_PUPPET_VERSION`, `facter` and `hiera` *do* get installed properly with the above 2 lines commented.  See the Puppet `.gemspec` at https://github.com/puppetlabs/puppet/blob/master/.gemspec#L41-L54

Results are similar to my test at https://github.com/puppetlabs/puppetlabs-registry/pull/85/files#r33734611

```
C:\source\puppetlabs-registry [pr/85]> ruby --version
ruby 2.1.5p273 (2014-11-13 revision 48405) [x64-mingw32]

C:\source\puppetlabs-registry [pr/85]> echo $env:GEM_PUPPET_VERSION
file://c:\source\puppet

C:\source\puppetlabs-registry [pr/85]> del gemfile.lock
C:\source\puppetlabs-registry [pr/85]> del .\.bundle\ -Recurse -Force

C:\source\puppetlabs-registry [pr/85]> bundle install --path .bundle\gems
DL is deprecated, please use Fiddle
Fetching gem metadata from https://rubygems.org/..........
Resolving dependencies...
Installing CFPropertyList 2.3.1
Installing i18n 0.7.0
Installing rake 10.4.2
Installing minitest 5.7.0
Installing addressable 2.3.8
Installing thread_safe 0.3.5
Installing extlib 0.9.16
Installing multi_json 1.11.1
Installing mini_portile 0.6.2
Installing excon 0.45.3
Installing builder 3.2.2
Installing formatador 0.2.5
Installing mime-types 2.6.1
Installing ipaddress 0.8.0
Installing net-ssh 2.9.2
Installing inflecto 0.0.2
Installing retriable 1.4.1
Installing multipart-post 2.0.0
Installing jwt 1.5.1
Installing json 1.8.3
Installing hocon 0.9.2
Installing open_uri_redirections 0.2.1
Installing inifile 2.0.2
Installing trollop 2.1.2
Installing rsync 1.0.9
Installing diff-lcs 1.2.5
Installing unf_ext 0.0.7.1
Installing json_pure 1.8.2
Installing metaclass 0.0.4
Installing puppet-lint 1.1.0
Installing rspec-core 2.14.8
Using bundler 1.7.12
Installing rspec-mocks 2.14.6
Installing fission 0.5.0
Installing puppet-syntax 2.0.0
Installing launchy 2.4.3
Installing autoparse 0.3.3
Installing tzinfo 1.2.2
Installing net-scp 1.2.1
Installing faraday 0.9.1
Installing nokogiri 1.6.6.2
Installing docker-api 1.21.4
Installing rspec-expectations 2.14.5
Installing unf 0.1.4
Installing mocha 0.10.5
Installing activesupport 4.2.3
Installing fog-core 1.31.1
Installing signet 0.6.1
Installing rbvmomi 1.8.2
Installing rspec 2.14.1
Installing aws-sdk-v1 1.64.0
Installing fog-xml 0.1.2
Installing fog-json 1.0.2
Installing fog-local 0.2.1
Installing fog-vmfusion 0.1.0
Installing google-api-client 0.8.2
Installing rspec-puppet 2.2.0
Installing aws-sdk 1.64.0
Installing fog-atmos 0.1.0
Installing fog-profitbricks 0.0.3
Installing fog-ecloud 0.3.0
Installing fog-terremark 0.1.0
Installing fog-voxel 0.1.0
Installing fog-brightbox 0.7.2
Installing fog-google 0.0.6
Installing fog-aws 0.6.0
Installing fog-powerdns 0.1.1
Installing fog-radosgw 0.0.4
Installing fog-riakcs 0.1.0
Installing fog-sakuracloud 1.0.1
Installing fog-serverlove 0.1.2
Installing fog-storm_on_demand 0.1.1
Installing fog-softlayer 0.4.7
Installing puppetlabs_spec_helper 0.10.3
Installing fog 1.31.0
Installing beaker 2.15.1
Installing beaker-puppet_install_helper 0.2.1
Installing ffi 1.9.10
Installing win32-dir 0.4.9
Installing win32-security 0.2.5
Installing hiera 2.0.0
Installing facter 2.4.4
Using puppet 4.1.0 from source at c:/source/puppet
Your bundle is complete!
It was installed into ./.bundle/gems
Post-install message from nokogiri:
Nokogiri is built with the packaged libraries: libxml2-2.9.2, libxslt-1.1.28, zlib-1.2.8, libiconv-1.14.
```

ferventcoder: `Installing facter 2.4.4`
versus `Using facter 2.4.4 from source at c:/source/facter`

Gemfile
ferventcoder: Kept around because we've had that on gemfiles in the past:
https://github.com/puppetlabs/puppetlabs-acl/blob/master/Gemfile#L39-L40

ferventcoder: It's probably a relic that can be removed.

Gemfile
ferventcoder: Yeah, looks like you are getting facter dependencies and not the others:

```
Installing win32-dir 0.4.9
Installing win32-security 0.2.5
Installing hiera 2.0.0
Installing facter 2.4.4
```

Gemfile
ferventcoder: Some folks fake out all three, but especially facter because multiple facters in the past caused issues with spec testing

Gemfile
ferventcoder: This is interesting to me -

```
C:\source\puppetlabs-registry [pr/85]> echo $env:GEM_PUPPET_VERSION
file://c:\source\puppet
```

I would have expected that to be a version. The way of faking out the version has been PUPPET_LOCATION in the past. Has that changed?

ferventcoder: This is what is happening when you fake out GEM_PUPPET_VERSION in that way:

```ruby
if ENV['PUPPET_LOCATION'] || (ENV['GEM_PUPPET_VERSION'] && ENV['GEM_PUPPET_VERSION'] < '3.5.0')
```
`ENV['GEM_PUPPET_VERSION'] < '3.5.0')` or:
```
'file://c:\source\puppet' < '3.5.0'
=> false
```

ferventcoder: I fixed this up, plus I noticed that the recommendation from modulesync is to use `PUPPET_GEM_VERSION` - https://github.com/puppetlabs/modulesync_configs/blob/master/moduleroot/Gemfile#L56

Gemfile
```diff
@@ -28,31 +47,64 @@ group :system_tests do
...
+    gem "windows-pr", "~> 1.2", :require => false
+  end
+
+  # If you are using file locations for Puppet it
+  # may not catch these two dependencies, so we
+  # will want those as well. Plus allow for
+  # overriding those locations as well with source
+  # locations.
+  gem "facter", *location_for( ENV['FACTER_GEM_VERSION'] || ENV['GEM_FACTER_VERSION'] || ENV['FACTER_LOCATION'] || '>= 1')
+  gem "hiera", *location_for(ENV['HIERA_GEM_VERSION'] || ENV['GEM_HIERA_VERSION'] || ENV['HIERA_LOCATION'] || '>= 1')
```
ferventcoder: These should not be in the explicit windows section.

ferventcoder:
@Iristyle @joshcooper this is ready - I talked to @cyberious and we are going to fix up travis.yml dependencies as part of modulesync.

Gemfile
```diff
@@ -28,31 +46,64 @@ group :system_tests do
...
-if puppetversion
-  gem 'puppet', *location_for(puppetversion)
-else
-  gem 'puppet', :require => false
+if explicitly_require_windows_gems
+  gem "ffi", "~> 1.9.0", :require => false
+  gem "win32-dir", "~> 0.3", :require => false
+  gem "win32-eventlog", "~> 0.5", :require => false
+  gem "win32-process", "~> 0.6", :require => false
+  gem "win32-security", "~> 0.1", :require => false
```
Iristyle: So it appears that this require is now pulling in old Facter versions when specifying `file://` as the Puppet source.

To recap on our historical inclusions:

* Puppet 4.2.0
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Puppet 4.1.0
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Puppet 4.0.0
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Puppet 3.8.1
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Puppet 3.7.5
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Puppet 3.7.0
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Puppet 3.6.2
    * x86 `~> 0.1.4`
* Puppet 3.5.1
    * x86 `~> 0.1.4`
* Puppet 3.4.3
    * N/A
* Puppet 3.3.2
    * N/A

For Facter we have

* Facter 2.4.4
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Facter 2.4.0
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Facter 2.3.0
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Facter 2.2.0
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Facter 2.1.0
    * x86 `~> 0.2.5`
    * x64 `~> 0.2.5`
* Facter 2.0.1
    * x86 - not included / used
* Facter 1.7.6 and BELOW
    * N/A


So I think we should be able to cook up something that works.

Gemfile
```diff
@@ -28,31 +46,64 @@ group :system_tests do
...
+  gem "minitar", "~> 0.5.4", :require => false
+
+  if RUBY_VERSION =~ /^1\./ then
+    gem "win32console", :require => false
+  end
+
+  # Puppet less than 3.7.0 requires these
+  # Granted that anything over 3.5.0 won't make it in
+  # here anyway, but at least the check is accurate.
+  if puppet_gem_location != :gem || puppetversion < '3.7.0'
```
ferventcoder: This override needs to also apply to source, which likely means we don't need it.

Gemfile
ferventcoder: So this should now be working with all scenarios.

Iristyle:
Hmm @ferventcoder, latest push da84dfb has some weird failures in Travis https://travis-ci.org/puppetlabs/puppetlabs-registry/builds/69802089

Gemfile
```diff
@@ -28,31 +46,60 @@ group :system_tests do
...
-      gem "win32console", :require => false
+# The recommendation is for PROJECT_GEM_VERSION,
+# there are older ways of referencing these. Add
+# them all for compatibility reasons. Remove
+# later when no issues are known
+puppetversion = ENV['PUPPET_GEM_VERSION'] || ENV['GEM_PUPPET_VERSION'] || ENV['PUPPET_LOCATION'] || '>= 0'
+gem 'puppet', *location_for(puppetversion)
+
+gem "facter", *location_for(ENV['FACTER_GEM_VERSION'] || ENV['GEM_FACTER_VERSION'] || ENV['FACTER_LOCATION'] || '>= 0')
+gem "hiera", *location_for(ENV['HIERA_GEM_VERSION'] || ENV['GEM_HIERA_VERSION'] || ENV['HIERA_LOCATION'] || '>= 0')
```
Iristyle: So the explicit inclusion of Hiera here is what runs afoul in Travis.  I turned on `DEBUG_RESOLVER=1` in a separate PR and I can see that Bundler has trouble with resolving Hiera from Puppet 3.8.1 all the way down, until Puppet 3.0.0, at which point it finally matches on a Hiera version - see https://api.travis-ci.org/jobs/69829416/log.txt?deansi=true

So yeah, Bundler--

ferventcoder: Puppet 3.0.0 in that TravisCI failure with Hiera 3.0.1. Seriously that's messed up.

ferventcoder: But also those nodes that are failing are going to be addressed with the travis ci fix. Should we go ahead and address those and then come back to this?

Iristyle:
I think that we're finally there with b9b7c6e - that basically has the same set of Gemfile changes now as #87 that I've been working on.  I think we should add a few more details to the commit message describing the scenarios we "officially" support if possible (and mention our one failure).  If you can think of anything else in addition to these, it would be great:

* We fully support specifying Puppet and Facter versions as environment variables, and Bundler should "do the right thing" when resolving dependencies
* These versions may be file:// for Puppet OR Facter OR both Puppet and Facter to support the scenario of making changes to Puppet in concert with local module changes
* Bundler behaves better when omitting a dependency that is not necessary to explicitly include.  For instance, adding Hiera >= 0, when it was not specified in an environment variable, resulted in some rather surprising Bundler behavior.  Namely it resolved an odd Hiera version, which caused a cascading effect of incorrect dependency resolution for other gems (namely win32-security).
* Right now, TravisCI exhibits some weird behavior with installing Facter 1.7.6 for Puppet ~> 3.0 (presently 3.8.1) because it appears to be looking at the universal-darwin Facter gem.  Beaker transitively requires CFPropertyList ~> 2.2, which installs 2.3.1.  This later conflicts with the ~> 2.2.6 CFPropertyList requirement that newer Facter versions have.  Bundler works backward to find a Facter gem that doesn't use CFPropertyList, namely 1.7.6
* Note that if you need to better understand Bundlers resolution mechanism, set the `DEBUG_RESOLVER` environment variable to 1

ferventcoder:
Add to the commit message and probably to the actual gemfile some of those details.

ferventcoder:
@Iristyle and here we go.

Iristyle:
:shipit:

PR #82, opened 2015-06-03
(maint) Modulesync updates of various parts

cyberious:
Adds Appveyor, Gemfile fixes and spec/spec.opts using modulesync

PR #88, opened 2015-07-10
(MODULES-2207) Update Modulesync

ferventcoder:
This ensures the following items are under moduleysnc:

.gitignore
.travis.yml
CONTRIBUTING.md
Gemfile
appveyor.yml
spec/spec.opts

Most importantly - this ensures that acl has up to date Gemfile, Travis CI and
AppVeyor configurations that allow it to be tested properly.

Supersedes #82


Gemfile
```diff
@@ -67,15 +68,8 @@ explicitly_require_windows_gems = false
 puppet_gem_location = gem_type(puppetversion)
 # This is not a perfect answer to the version check
 if puppet_gem_location != :gem || puppetversion < '3.5.0'
-  is_x64 = Gem::Platform.local.cpu == 'x64'
-  if is_x64
-    platform(:x64_mingw) do
-      explicitly_require_windows_gems = true
```
ferventcoder: Apparently this is always evaluated to true. Ugh.

.project
```diff
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <projectDescription>
-	<name>registry</name>
+	<name>puppetlabs-registry</name>
```
cyberious: Do we want to rename the projects?

ferventcoder: I didn't change your .project settings

ferventcoder: So you tell me haha

cyberious: Might be something in modsync itself

Gemfile
```diff
@@ -25,22 +25,22 @@ def location_for(place_or_version, fake_version = nil)
...
 end
-
-beaker_version = ENV['BEAKER_VERSION'] || '~> 2.2'
 group :system_tests do
+  beaker_version = ENV['BEAKER_VERSION'] || '~> 2.2'
   if beaker_version
     gem 'beaker', *location_for(beaker_version)
   else
-    gem 'beaker',                :require => false, :platforms => :ruby
+    gem 'beaker', :require => false, :platforms => :ruby
```
Iristyle: It seems that this code path is not necessary, is it @cyberious ?  To reach it, I need to intentionally set `ENV['BEAKER_VERSION']` to an empty string prior to installing gems, since `beaker_version` defaults to `~> 2.2` when it's not set..

Is this something we need to support for some reason in Linux / OSX?  (With `:platforms => :ruby` I don't believe that beaker will be installed ever on JRuby or Windows)

/cc @ferventcoder

Gemfile
```diff
@@ -67,15 +66,8 @@ explicitly_require_windows_gems = false
...
-  is_x64 = Gem::Platform.local.cpu == 'x64'
-  if is_x64
-    platform(:x64_mingw) do
-      explicitly_require_windows_gems = true
-    end
-  else
-    platform (:mingw) do
-      explicitly_require_windows_gems = true
-    end
+  if Gem::Platform.local.os == 'mingw32'
```
Iristyle: So previously we intended to explicitly required gems on Windows if:

* Running a 64-bit Ruby on a 64-bit Windows OS
* Running a 32-bit Ruby on a 32-bit Windows OS
* But not if we're running a 32-bit Ruby on a 64-bit OS

With this change, we will explicitly require gems on Windows all the time.  

It seems that this actually fixes a problem in our prior logic - am I understanding that correctly? @ferventcoder I believe this was the issue you saw in TravisCI --  Puppet < 3.5 + a 32-bit Ruby 1.9.3 installed to a 64-bit OS?

:+1:

Gemfile
ferventcoder: I see what you are stating. That would be something for cleanup, yes.

Gemfile
ferventcoder: The intention was to always require gems on Windows all the time if we are overriding the Puppet Gem location or using a version less than 3.5.0. The old logic had a bug in that it would also include non-Windows plaforms if either of the other two concerns were met. The fix is https://github.com/ferventcoder/modulesync_configs/commit/518401d8f494b1b1f82754b48278e9bd13d2d315

ferventcoder: And just realized my spelling error there. :/

ferventcoder: actually not a spelling error, but fixing up the commit anyway.

Gemfile
ferventcoder: https://github.com/ferventcoder/modulesync_configs/commit/fd5d94586f4c64a4ee1f473c25a65b6e26be4c5f#commitcomment-12183398

PR #89, opened 2015-07-15
(MODULES-2207) Gem restrictions for Older puppet versions

ferventcoder:
Further restrict down the restrictions on windows dependencies when
using older Puppet versions, such that it works with Puppet 3.0.0 -
3.7.0. It does require a bit more logic in selecting gems but it allows
for testing older platforms without any additional known issues for
Puppet versions under 3.4.0.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -10,7 +10,7 @@ Gemfile:
     ':development':
       - gem: rake
       - gem: rspec
-        version: '~>2.14.0'
+        version: '~>2.14.1'
       - gem: puppet-lint
       - gem: puppetlabs_spec_helper
       - gem: puppet_facts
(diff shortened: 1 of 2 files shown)


.sync.yml
```diff
@@ -18,6 +18,7 @@ Gemfile:
         version: '~>0.10.5'
     ':system_tests':
       - gem: beaker
+      - gem: beaker-rspec
```
cyberious: This should not be here, we should put this in sync.yml when needed.

ferventcoder: I'm 6 to one, half dozen to the other on this. Was more for supporting what @justinstoller might be calling in his scripts. Does it do harm if it is there?

cyberious:
@ferventcoder can we clean up the commits of add and remove beaker-rspec

ferventcoder:
Yes.

ferventcoder:
Removed.

PR #94, opened 2015-08-12
(maint) add environment info to acceptance tests

ferventcoder:
Report on environment information so we know what we expected is
installed during the setup phase.


Iristyle:
This looks like something that should be a part of modulesync ... thoughts?

ferventcoder:
I was originally thinking it should be part of beaker-puppet_install_helper.

PR #98, opened 2015-11-05
(FM-4917) Rake / Beaker / Rspec compatibility

ferventcoder:
Beaker needs a version of Rake less than 11 due to using
`last_comment' in rake_task.rb:62. Reduce the version of
Rake down to the same dependency that Beaker does as a
development dependency.

Rspec 3+ requires that should be explicitly set as an expect term.
be_true and be_false have also been changed so that they are not
whether something is truthy or falsey, but simply really true/false. To
get the older behavior, the test simply needs to check for be_truthy or
be_falsey. For explicit true/false checks, it is be true and be false.

Remove gem install bundler as part of modulesync updates.

Paired with Glenn Sarti <glenn.sarti at puppetlabs dot com>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,16 +25,28 @@ def location_for(place_or_version, fake_version = nil)
   end
 end
 
+# The following gems are not included by default as they require DevKit on Windows.
+# You should probably include them in a Gemfile.local or a ~/.gemfile
+#gem 'pry' #this may already be included in the gemfile
+#gem 'pry-stack_explorer', :require => false
+#if RUBY_VERSION =~ /^2/
+#  gem 'pry-byebug'
+#else
+#  gem 'pry-debugger'
+#end
+
 group :development do
   gem 'rake',                                :require => false
-  gem 'rspec', '~>2.14.1',                   :require => false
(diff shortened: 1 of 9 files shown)


tphoney:
@ferventcoder there is an error. so close :)

ferventcoder:
That's an interesting issue - I will check out the details and fix it - but first, retry appveyor. :)

Gemfile
```diff
@@ -25,16 +25,28 @@ def location_for(place_or_version, fake_version = nil)
...
+#if RUBY_VERSION =~ /^2/
+#  gem 'pry-byebug'
+#else
+#  gem 'pry-debugger'
+#end
+
 group :development do
   gem 'rake',                                :require => false
-  gem 'rspec', '~>2.14.1',                   :require => false
+  gem 'rspec', '~>3.0',                      :require => false
```
ferventcoder: Upgrade to new rspec may cause failures.

ferventcoder:
@puppetlabs/windows FIX the world.

ferventcoder:
@glennsarti okay, this should be super helpful this time. Let's see how Travis does.

PR #102, opened 2016-03-24
(maint) Fix the style of the registry::value parameters

natemccurdy:
Prior to this, the parameters in the defined type, registry::value, were
all on one line and not spaced according to the Puppet Labs style guide.

This commit fixes that by putting each parameter on its own line and
lining everything up.


ferventcoder:
Thanks!

PR #107, opened 2016-05-19
(FM-5235) normalize case for $key strings

binford2k:
Before this, if someone managed a value and wasn't consistent in the
case they used for `$key`, then it would fail with a duplicate resource.

### DIFF
--- a/manifests/value.pp
+++ b/manifests/value.pp
@@ -57,7 +57,8 @@ define registry::value (
   validate_re($type, '^\w+',
     'type parameter must not be empty but it is type => "$type"')
 
-
+  # normalize the case so that our defined() check is more consistent
+  $key_normalized = downcase($key)
 
   $value_real = $value ? {
     undef       => $name,
@@ -69,13 +70,13 @@ define registry::value (
   Registry_key { ensure => present }
   Registry_value { ensure => present }
 
-  if !defined(Registry_key[$key]) {
-    registry_key { $key: }
+  if !defined(Registry_key[$key_normalized]) {
+    registry_key { $key_normalized: }
   }
 
   # If value_real is an empty string then the default value of the key will be
   # managed.
-  registry_value { "${key}\\${value_real}":
+  registry_value { "${key_normalized}\\${value_real}":
     type => $type,
     data => $data,
   }


manifests/value.pp
```diff
@@ -57,7 +57,8 @@
   validate_re($type, '^\w+',
     'type parameter must not be empty but it is type => "$type"')
 
-
+  # normalize the case so that our defined() check is more consistent
+  $key_normalized = downcase($key)
```
jpogran: Wouldn't this be a surprise to a user? While the registry key is case insensitive, what if the user wanted some or all of it to be upper case? Here we would pass the lower case text down to the provider to be created as lower case

binford2k: In that case, they can use a `registry_value` resource directly, I guess? IMHO, normalizing to a known case is less of a surprise than arbitrarily breaking when they use inconsistent casing, since they're used to having that freedom.

jpogran: @puppetlabs/windows thoughts?

ferventcoder: The only issue is in setting the value. It should carry the case that the user sets the resource with when the manifest results in creating the resource. Other than that, 👍

Iristyle: It's an interesting proposition - comparisons are insensitive, but setting values is case preserving.

If the key matches case insensitively, then it implies values should not be modified by Puppet.  But I agree with @ferventcoder that when the value *is* being set, it should apply the case that the user set in their manifest.

manifests/value.pp
```diff
@@ -69,13 +70,13 @@
   Registry_key { ensure => present }
   Registry_value { ensure => present }
 
-  if !defined(Registry_key[$key]) {
-    registry_key { $key: }
+  if !defined(Registry_key[$key_normalized]) {
+    registry_key { $key_normalized: }
```
ferventcoder: This should probably be the original

binford2k: It it defines a `registry_value` with the original case, there's no way to reliably check to see if it's already defined, unless we write a new `case_insensitive_defined()` function.

lionce:
Hello @binford2k,

Please check again this PR. It seems like now it has some conflicts If you still want this to be merged, please rebase it with the latest code.

Thanks

sheenaajay:
Hi @binford2k  ,Hasn't seen any activity in this PR, do you think you'll be able to work on it in the near future? No pressure, just checking in.Thank you.Please let us know if you need more information.

binford2k:
Hey @sheenaajay! I honestly don't know what the right thing to do here is. I'll happily resolve conflicts, but I think that @Iristyle should decide correct behavior.

michaeltlombardi:
The original zendesk ticket that drove this change is closed. Regarding correct behavior, this is likely to be a breaking change for most current users, which is something we may not want to do given our current resource restraints and the lack of new calls for this functionality.

I'm closing this for now - if, in the future, someone has the time and bandwidth to tackle this or additional customers raise the issue, we can/will revisit.

PR #114, opened 2016-09-12
(MODULES-3865) Add support for for hkey_users

hilhorst:
I took the less controversial parts from #80 and made a new pull request.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -217,7 +217,7 @@ Puppet's [native service resource](http://docs.puppet.com/references/latest/type
 
 ##Limitations
 
-* Keys within HKEY_LOCAL_MACHINE (hklm) or HKEY_CLASSES_ROOT (hkcr) are supported. Other predefined root keys (e.g., HKEY_USERS) are not currently supported.
+* Keys within HKEY_LOCAL_MACHINE (hklm), HKEY_CLASSES_ROOT (hkcr) or HKEY_USERS (hku) are supported. Other predefined root keys (e.g., HKEY_CURRENT_USER) are not currently supported.
 * Puppet doesn't recursively delete Registry keys.
 
 Please report any issues through our [Module Issue Tracker](https://tickets.puppet.com/browse/MODULES).
--- a/lib/puppet_x/puppetlabs/registry.rb
+++ b/lib/puppet_x/puppetlabs/registry.rb
@@ -95,8 +95,9 @@ module Registry
                 :hklm
               when /hkey_classes_root/, /hkcr/
                 :hkcr
+              when /hkey_users/, /hku/
+                :hku
               when /hkey_current_user/, /hkcu/,
-                /hkey_users/, /hku/,
                 /hkey_current_config/, /hkcc/,
                 /hkey_performance_data/,
                 /hkey_performance_text/,
--- a/lib/puppet_x/puppetlabs/registry/provider_base.rb
+++ b/lib/puppet_x/puppetlabs/registry/provider_base.rb
@@ -141,6 +141,7 @@ module ProviderBase
         {
           :hkcr => Win32::Registry::HKEY_CLASSES_ROOT,
           :hklm => Win32::Registry::HKEY_LOCAL_MACHINE,
+          :hku  => Win32::Registry::HKEY_USERS,
         }
       end
     end
--- a/spec/unit/puppet/type/registry_key_spec.rb
+++ b/spec/unit/puppet/type/registry_key_spec.rb
@@ -39,6 +39,12 @@ describe Puppet::Type.type(:registry_key) do
         key[:path] = path
       end
     end
+    
+    %w[hku hku\.DEFAULT hku\.DEFAULT\software hku\.DEFAULT\software\vendor].each do |path|
+      it "should accept #{path}" do
+        key[:path] = path
+      end
+    end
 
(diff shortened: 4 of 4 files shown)


ferventcoder:
Hello @hilhorst - here's what we really want to do. Take a look at https://github.com/badgerious/puppet-windows-env/blob/2d11f3021d08418670f913cc8f2da7a814c3d2c3/lib/puppet/provider/windows_env/windows_env.rb#L303-L327 to understand how to load the user hive.

glennsarti:
HKEY_USERS is still a fair target for this module to use though.  Particularly for HKU\.DEFAULT.

Loading external hives is very different task.

hilhorst:
Thanks for the pointer, @ferventcoder. I'll certainly have a go at it but there is very little working in my advantage. This is my first time working on a puppet module, the first time programming in Ruby even, and I have little to no knowledge of the windows registry. But let's just say I'm up for a challenge 😉.

@glennsarti, that was what I was going for. Not for managing everything in the user hives, but just the defaults under HKEY_USERS. A colleague of mine needed just that for managing part of our security baseline through Puppet.

hilhorst:
@ferventcoder, I'm trying to figure out what the resource would look like when managing a key and/or value for a particular user.

I guess you'd want to specify a user, because I don't see how the 'current user' makes sense in the context of managing the desired state of a server with Puppet, where the agent is run once every half hour. From what I understand hkcu is just a shortcut to a user under hku anyways.

When a user's hive is loaded under hkey_users the path contains the sid for this user, but I guess you wouldn't want to use the path with this sid in your resource definition, because you might not yet know the sid of the user when you're defining the resource.

I see some alternative solutions:
1. Use the path parameter but with the username instead of the sid 
eg. (:path => 'hku\[user "domain\user"]\environment')
2. Use a separate (optional) parameter for specifying the user, and just insert it's corresponding sid into the given path
eg. (:path => 'hku\environment' , :user => 'domain\user')
3. Use a separate (optional) parameter for specifying the user, with a placeholder in the path
eg. (:path => 'hku\[user]\environment' , :user => 'domain\user')

Any thoughts or suggestions?

glennsarti:
> When a user's hive is loaded under hkey_users the path contains the sid for this user, but I guess you wouldn't want to use the path with this sid in your resource definition

Not true, that's just how Windows named it.  You can load a hive with an arbitrary name.  I would _not_ recommend mounting a hive with it's SID name just in case.

ferventcoder:
I can see HKU. Out of curiosity, how does HKU change existing users?

ferventcoder:
I've always seen managing a user based on user name or SID, where it loads the user hive and then makes changes. It's very similar to how the windows_env module does it (not sure if you explored that link further to see how the resources are defined).

Similar to option 2 or 3 above. You may be specifying hkcu and then user is required.

glennsarti:
I would suggest moving the conversation about user loads etc. to either a Jira ticket or GH issue.  That way we can concentrate on merging in the changes for HKU support and then address the loading behavior later.

ferventcoder:
@glennsarti good plan - thought I already created an issue on that.

ferventcoder:
Apologies, I was looking at HKU and for a particular user as the same thing.

glennsarti:
https://tickets.puppetlabs.com/browse/MODULES-422 is for loading hives etc.

ferventcoder:
MODULES-422 is/was the JIRA ticket for HKU. Would you mind updating your commit messages to reflect this @hilhorst?

ferventcoder:
@glennsarti now I can see why I was confused - if MODULES-422 is not the same thing as the changes proposed here, we should fix up that ticket (starting with the title).

glennsarti:
MODULES-422 is for adding the ability to mount./dismount reg hives etc for users.  This PR is about being able to add/remove keys etc. inside HKU.

glennsarti:
I don't believe there's a ticket logged for this

hilhorst:
Can we maybe merge this without a ticket? :grin:

ferventcoder:
Actually, let's create a new ticket for it. We'll need a ticket to point to for changes that have occurred and changelogs.

glennsarti:
Raise ticket https://tickets.puppetlabs.com/browse/MODULES-3865

glennsarti:
@jpogran @ferventcoder This PR looks good to go.  Can we get a merge on it?

wilson208:
@ferventcoder @jpogran Any issues with merging this and updating the related ticket?

ferventcoder:
Yes, this looks good. Just needs the commits updated to include the JIRA ticket in the commit summary. @hilhorst did you want us to take care of that on merge or did you want to do that as part of this PR?

hilhorst:
@ferventcoder If you could do that on merge, that would be great. I'm afraid I'll mess things up with the forced push.

ferventcoder:
Okay. Will handle.

ferventcoder:
Rebased and merged into master at https://github.com/puppetlabs/puppetlabs-registry/commit/c515df48c5c0f3e91fb732d28d334c8e6fe15934. Thanks for your contribution @hilhorst!

hilhorst:
Thanks @ferventcoder and @glennsarti, for your guidance. It was very nice to be able to contribute and I look forward to possibly contribute more in the future :)

PR #117, opened 2016-11-16
Workaround frozen strings on ruby 1.9

hunner:
https://github.com/puppetlabs/modulesync_configs/pull/104


ferventcoder:
Are all of these ready to go or do they need any more updates?

PR #123, opened 2017-01-24
(MODULES-4331) quoted UNSET string

binford2k:
This is no longer a valid bareword.


ferventcoder:
Ruh roh. Looks good, thanks for that!

PR #124, opened 2017-02-08
Add backslash support to registry resource

sbrito85:
Puppet bug here: https://tickets.puppetlabs.com/browse/MODULES-2957

Picked up the work being done on other pull requests.

### DIFF
--- a/lib/puppet/type/registry_key.rb
+++ b/lib/puppet/type/registry_key.rb
@@ -103,7 +103,7 @@ EOT
 
     # get the "should" names of registry values associated with this key
     should_values = catalog.relationship_graph.direct_dependents_of(self).select {|dep| dep.type == :registry_value }.map do |reg|
-      PuppetX::Puppetlabs::Registry::RegistryValuePath.new(reg.parameter(:path).value).valuename
+      PuppetX::Puppetlabs::Registry::RegistryValuePath.new(reg.parameter(:path).value, reg.parameter(:value_name).value).valuename
     end
 
     # get the "is" names of registry values associated with this key
--- a/lib/puppet/type/registry_key.rb
+++ b/lib/puppet/type/registry_key.rb
@@ -36,10 +36,10 @@ EOT
       prefix.  For example: '32:HKLM\Software'"
 
     validate do |path|
-      PuppetX::Puppetlabs::Registry::RegistryKeyPath.new(path, '').valid?
+      PuppetX::Puppetlabs::Registry::RegistryKeyPath.new(path).valid?
     end
     munge do |path|
-      reg_path = PuppetX::Puppetlabs::Registry::RegistryKeyPath.new(path, '')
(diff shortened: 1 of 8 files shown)


sbrito85:
Looks like I need to clean up some tests :)

jpogran:
@sbrito85 thanks for the PR! I've added this to triage and it will be reviewed this week. While that happens, could you review the [CONTRIBUTING.md](https://github.com/puppetlabs/puppetlabs-registry/blob/master/CONTRIBUTING.md) file and update your commits? Specifically referring to commit message section.

lib/puppet/type/registry_key.rb
```diff
@@ -103,7 +103,7 @@ def eval_generate
 
     # get the "should" names of registry values associated with this key
     should_values = catalog.relationship_graph.direct_dependents_of(self).select {|dep| dep.type == :registry_value }.map do |reg|
-      PuppetX::Puppetlabs::Registry::RegistryValuePath.new(reg.parameter(:path).value).valuename
+      PuppetX::Puppetlabs::Registry::RegistryValuePath.new(reg.parameter(:path).value, reg.parameter(:value_name).value).valuename
```
jpogran: Does this mean it's a breaking change? As in, do users using the old syntax have to use this new syntax?

sbrito85: Old syntax still works from all tests that I've done. This should not be a breaking change.

glennsarti:
Hi @sbrito85.  Thanks for all your work on this so far.  I have some concerns about it;

* Needs documentation update to explain how to use a value with a name with '\`
* Needs spec tests for valuenames with a `\`
* I briefly started adding some tests and noticed there are some issues with the auto-require logic as it doesn't check for resources with `value_name`

sbrito85:
I plan on having some cycles this week to look at these concerns. Sorry for the delay!

1) Will add documentation, should have that in today or tomorrow
2) For this part, I may need some assistance. 
3) Where should I be checking for this? I presume it should go in lib/puppet/type/registry_value.rb like the path value?

glennsarti:
@sbrito85 

I started work on squashing your commits, and then adding tests.

Commits are here;
https://github.com/glennsarti/puppetlabs-registry/commits/ticket/master/MODULES-2957-backslash

This is what tripped me up with autorequire as those tests were failing.

The autorequire code is here
https://github.com/puppetlabs/puppetlabs-registry/blob/master/lib/puppet/type/registry_value.rb#L124-L137

Note it's using the resource name not the non-backslash value.

michaeljmatthews22:
Any update on this? Really looking forward to this being available for our company.

makomatic:
Any Update? This would reallly be nice...

Iristyle:
This is being superseded in #143 and #142 - as such, closing.

Thanks for the original PR @sbrito85!

PR #138, opened 2017-11-07
(maint) Modify beaker rake task to invoke beaker directly

glennsarti:
The acceptance tests for Registry are written in standard beaker format however
the preferred method is using beaker-rspec.  This commit overrides the
default `beaker` task, which would normally use beaker-rspec, and instead
invokes beaker directly.  This is only need while the module tests are migrated
to the newer rspec-beaker format.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -25,6 +25,19 @@ end
 
 task :default => [:test]
 
+# The acceptance tests for Registry are written in standard beaker format however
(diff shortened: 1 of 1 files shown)


glennsarti:
Blocked - DO NOT MERGE.

glennsarti:
Adhoc has passed with success.  Ready for merge.

PR #139, opened 2017-11-14
(MODULES-5976) convert beaker tests to beaker rspec tests

ThoughtCrhyme:


Iristyle:
Could you provide the invocation that you're using for the old code vs the new code?

spec/acceptance/nodesets/default.yml
```diff
@@ -1,18 +1,24 @@
...
-  windows2012-64-1:
-    pe_dir:
-    pe_ver:
-    pe_upgrade_dir:
-    pe_upgrade_ver:
-    hypervisor: vmpooler
-    platform: windows-2012-64
-    ruby_arch: x64
-    template: win-2012-x86_64
+  centos7-master:
```
Iristyle: Do we want a master in the default nodeset here?  Could be a commit after this work...

spec/acceptance/should_create_key_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+
+  registry_key { '#{keypath}\\SubKeyToPurge': }
+  if $architecture == 'x64' {
+    registry_key { '32:#{keypath}\\SubKeyToPurge': }
+  }
+PHASE2
+
+    # Delete our keys
+    phase3 = <<PHASE3
+  registry_key { 'HKLM\\Software\\Vendor': ensure => present }
```
glennsarti: Not sure if this is super important, but I'm seeing odd whitespace and indenting here.

spec/spec_helper_acceptance.rb
```diff
@@ -0,0 +1,59 @@
...
+  proj_root = File.expand_path(File.join(File.dirname(__FILE__), '..'))
+  hosts.each do |host|
+    if host['platform'] =~ /windows/
+      on host, "mkdir -p #{host['distmoduledir']}/registry"
+      target = (on host, "echo #{host['distmoduledir']}/registry").raw_output.chomp
+
+      %w(lib manifests metadata.json).each do |file|
+        scp_to host, "#{proj_root}/#{file}", target
+      end
+      on host, 'curl -k -o c:/puppetlabs-stdlib-4.6.0.tar.gz https://forgeapi.puppetlabs.com/v3/files/puppetlabs-stdlib-4.6.0.tar.gz'
```
glennsarti: I wonder if there are module_install_helpers for beaker we could use here instead?

e.g. https://github.com/puppetlabs/beaker-module_install_helper#install_module_dependencies_on

Side note - 4.6.0 is a REALLY old version

Rakefile
```diff
@@ -23,20 +23,20 @@ end
...
-# invokes beaker directly.  This is only need while the module tests are migrated
-# to the newer rspec-beaker format
-task_exists = Rake.application.tasks.any? { |t| t.name == 'beaker' }
-Rake::Task['beaker'].clear if task_exists
-desc 'Run acceptance testing shim'
-task :beaker do |t, args|
-  beaker_cmd = "beaker --options-file acceptance/.beaker-pe.cfg --hosts #{ENV['BEAKER_setfile']} --tests acceptance/tests --keyfile #{ENV['BEAKER_keyfile']}"
-  Kernel.system( beaker_cmd )
-end
+# task :default => [:test]
```
glennsarti: Can delete these instead of commenting.  Also could revert https://github.com/puppetlabs/puppetlabs-registry/commit/a168bd45db3ed492aace01b48d385b52fdd4f1c5

spec/acceptance/nodesets/default.yml
glennsarti: This file is coming from modsync so it probably shouldn't be modified.

spec/spec_helper_acceptance.rb
Iristyle: Yeah, I was wondering the same - seems like there should be a helper for these bits.

Rakefile
ThoughtCrhyme: Yup, forgot about that!

spec/spec_helper_acceptance.rb
ThoughtCrhyme: I can add this.

ThoughtCrhyme:
@Iristyle old invocation was
`bundle exec rake beaker:default`

Current invocation is the same, but add the environment variable of
`export PUPPET_INSTALL_TYPE=agent` to avoid testing with a master.

acceptance/config/masterless-windows-2008-x86_64.cfg
```diff
@@ -1,17 +0,0 @@
-HOSTS:
```
Iristyle: Can we remove these old configs in a separate commit?

Are they used by CI?

spec/acceptance/nodesets/default.yml
Iristyle: Also - is this why `sql_version: 2012` appears in here?

acceptance/config/masterless-windows-2008-x86_64.cfg
glennsarti: None of the nodeset is used in CI. All generated on the fly with BHG.

glennsarti:
Adhoc is green. Same tests, similar run times.  Bonus points for Jenkins now reading the test output XML files. It's merging time!

PR #142, opened 2017-12-12
{WIP}(MODULES-2957) Allow backslashes in registry value names

glennsarti:
Note - Builds on PR #143

Previously the registry_value resource could not update values with a backslash
in their name.  This was due to using the path as a way of expressing a unique
registry value, however values can only be uniquely named using a path and name.

This means the registry_value type needed a new parameter called `value_name`
which is a namevar.  Fortunately registry keys can not have a backslash in their
name so title_patterns could be used to split a resource title into it's path
and value name components.  This meant the type could maintain a level of
backwards compatibility.

This meant however that many of the assumptions about the :path parameter were
no longer true;
* It no longer contained the registry value_name (that is now in a separate
  param) so canonicalization is now the same as a registry_key
* Trailing backslashes needed to be deleted instead of kept
* Almost all of the custom code for the RegistryValuePath type was now redundant
* In most of the test cases, :path was used but this is no longer enough to
  identify a unique reg key value (It needs a Path and Value), however this
  could simply be changed to :title and then the title_pattern matching would
  extract the path and value_name.  This could also verify the backwards
  compatibiltiy behaviour of using a title

This commit also updates the unit tests for the new value_name parameter.

Unfortunately this is a breaking change even with the title_patterns.  Consider
the following:
```puppet
  registry_value { 'atitle\value':
    path => 'HKLM\Software\foo',
  }
```
Previously this would manage the value called 'foo' in the key 'HKLM\Software'.
With this change it would manage the value called 'value' (extracted from the
title_pattern because value_name is not specified) in the key 'HKLM\Software\foo'

```puppet
  registry_value { 'atitle':
    path => 'HKLM\Software\foo',
  }
```
Previously this would manage the value called 'foo' in the key 'HKLM\Software'.
With this change it would manage the default value (extracted from the
title_pattern because value_name is not specified) in the key 'HKLM\Software\foo'

- [x] Get it to work
- [x] Add tests
- [x] Confirm munging is no longer required
- [x] Add acceptance tests
- [x] Update documentation
- [ ] Test puppet resource

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -9,9 +9,8 @@
     * [Beginning with registry](#beginning-with-registry)
 4. [Usage - Configuration options and additional functionality](#usage)
 5. [Reference](#reference)
-    * [Public Defines](#public-defines)
-    * [Public Types](#public-types)
-    * [Parameters](#parameters)
+    * [Defined types](#defined-types)
+    * [Types/Providers](#typesproviders)
 6. [Limitations](#limitations)
 7. [Development - Guide for contributing to registry](#development)
 
@@ -51,13 +50,25 @@ registry_value { 'HKLM\System\CurrentControlSet\Services\Puppet\Description':
 }
 ```
 
+### Manage a single Registry value with a complex name
+
+``` puppet
+registry_value { 'ComplexValue':
+  path       => 'HKLM\System\CurrentControlSet\Services\Puppet',
+  value_name => 'ValueWithA\Backslash',
+  ensure     => present,
+  type       => string,
+  data       => "The Puppet Agent service periodically manages your configuration",
+}
+```
+
 ### Manage a Registry value and its parent key in one declaration
 
 ``` puppet
 class myapp {
   registry::value { 'puppetmaster':
-  key  => 'HKLM\Software\Vendor\PuppetLabs',
-  data => 'puppet.puppet.com',
+    key  => 'HKLM\Software\Vendor\PuppetLabs',
+    data => 'puppet.puppet.com',
   }
 }
 ```
@@ -139,17 +150,13 @@ registry::service { puppet:
 
 ## Reference
 
-### Public Defines
-* `registry::value`: Manages the parent key for a particular value. If the parent key doesn't exist, Puppet automatically creates it.
-* `registry::service`: Manages entries in the Microsoft service control framework by manipulating values in the key `HKLM\System\CurrentControlSet\Services\$name\`.
+### Defined types
 
-### Public Types
-* `registry_key`: Manages individual Registry keys.
-* `registry_value`: Manages individual Registry values.
+#### `registry::value`
 
-### Parameters
+Manages the parent key for a particular value. If the parent key doesn't exist, Puppet automatically creates it.
 
-#### `registry::value`:
+**Parameters**
 
 All parameters are required unless otherwise stated.
 
@@ -169,45 +176,13 @@ Provides the contents of the specified value. Valid options: a string by default
 
 *Optional.* Determines what Registry value(s) to manage within the specified key. To set a Registry value as the default value for its parent key, name the value '(default)'. Valid options: a string. Default value: the title of your declared resource.
 
-#### `registry_key`
-
-##### `ensure`
-
-Tells Puppet whether the key should or shouldn't exist. Valid options: 'present' and 'absent'. Default value: 'present'.
-
-##### `path`
-
-Specifies a Registry key for Puppet to manage. If any of the parent keys in the path don't exist, Puppet creates them automatically. Valid options: a string containing a Registry path. For example: `HKLM\Software` or `HKEY_LOCAL_MACHINE\Software\Vendor`.
-
-If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicitly managed using a prefix. For example: `32:HKLM\Software`.
-
-##### `purge_values`
-
-*Optional.* Specifies whether to delete any values in the specified key that are not managed by Puppet. Valid options: `true` and `false`. Default value: `false`.
-
-For more on this parameter, see the [Purge existing values section](#purge-existing-values) under Usage.
-
-#### `registry_value`
-
-##### `path`
-
-*Optional.* Specifies a Registry value for Puppet to manage. Valid options: a string containing a Registry path. If any of the parent keys in the path don't exist, Puppet creates them automatically. For example: `HKLM\Software` or `HKEY_LOCAL_MACHINE\Software\Vendor`. Default value: the title of your declared resource.
-
-If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicitly managed using a prefix. For example: `32:HKLM\Software\Value3`.
-
-##### `ensure`
-
-Tells Puppet whether the value should or shouldn't exist. Valid options: 'present' and 'absent'. Default value: 'present'.
-
-##### `type`
-
-*Optional.* Sets the data type of the specified value. Valid options: 'string', 'array', 'dword', 'qword', 'binary' and 'expand'. Default value: 'string'.
-
-##### `data`
-
-Provides the contents of the specified value. Valid options: a string by default; an array if specified through the `type` parameter.
-
 #### `registry::service`
+
+Manages entries in the Microsoft service control framework by manipulating values in the key `HKLM\System\CurrentControlSet\Services\$name\`.
+
+**Parameters**
+
+All parameters are required unless otherwise stated.
 
 ##### `ensure`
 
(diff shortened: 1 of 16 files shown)


glennsarti:
@Iristyle This is ready for an initial review.  There's still some code that is commented out.  That will be removed prior to this PR being ready for merge.

lib/puppet/type/registry_key.rb
```diff
@@ -112,7 +112,9 @@ def eval_generate
     # create absent registry_value resources for the complement
     resources = []
     (is_values - should_values).each do |name|
-      resources << Puppet::Type.type(:registry_value).new(:path => "#{self[:path]}\\#{name}", :ensure => :absent, :catalog => catalog)
+      # Registry values have a separate Path and Value Name, so just concatenate a resource title
+      title = "#{self[:path]}\\" + name.gsub('\\','')
```
glennsarti: Should be a unique name..

glennsarti:
Running though adhoc.

glennsarti:
Running though adhoc again.

glennsarti:
Passed Adhoc!

glennsarti:
Docs have been updated.  @clairecadman Would appreciate a review from you as my docs-fu needs improving.

README.markdown
```diff
@@ -187,49 +233,76 @@ If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicit
...
 For more on this parameter, see the [Purge existing values section](#purge-existing-values) under Usage.
 
+
 #### `registry_value`
 
-##### `path`
+Manages individual Registry values.
 
-*Optional.* Specifies a Registry value for Puppet to manage. Valid options: a string containing a Registry path. If any of the parent keys in the path don't exist, Puppet creates them automatically. For example: `HKLM\Software` or `HKEY_LOCAL_MACHINE\Software\Vendor`. Default value: the title of your declared resource.
+The title of the resource can determine the path and value name that will be managed, for example:
```
clairecadman: Suggestion: omit can. "The title of the resource determines the..."

README.markdown
```diff
@@ -187,49 +233,76 @@ If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicit
...
 
-If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicitly managed using a prefix. For example: `32:HKLM\Software\Value3`.
+The manifest below would manage the value called 'foo' in the key 'HKLM\Software'
```
clairecadman: Suggestion: omit would. "The manifest below manages the..."

README.markdown
```diff
@@ -187,49 +233,76 @@ If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicit
...
 
-##### `ensure`
+``` puppet
+registry_value { 'HKLM\Software\foo':
+  ...
+}
+```
 
-Tells Puppet whether the value should or shouldn't exist. Valid options: 'present' and 'absent'. Default value: 'present'.
+You can also manage the default value of a key by adding a backslash at the end of the path.  The manifest below would manage the default value in the key 'HKLM\Software\Vendor'
```
clairecadman: Suggestion: omit would. "The manifest below manages the..."

README.markdown
```diff
@@ -187,49 +233,76 @@ If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicit
...
 
-##### `type`
+``` puppet
+registry_value { 'HKLM\Software\Vendor\\':
+  ...
+}
+```
 
-*Optional.* Sets the data type of the specified value. Valid options: 'string', 'array', 'dword', 'qword', 'binary' and 'expand'. Default value: 'string'.
+Note that when explicitly setting a `path` or `value_name` parameter it is suggested to use both to avoid unexpected behavior, for example:
```
clairecadman: Suggestion: omit it is suggested - "Note that when explicitly setting a `path` or `value_name` parameter, use both to avoid unexpected behavior..."

README.markdown
```diff
@@ -187,49 +233,76 @@ If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicit
...
 
-##### `data`
+The manifest below would manage the value called 'value' (extracted from the title because value_name is not specified) in the key 'HKLM\Software\foo'
```
clairecadman: Suggestion: omit would. "The manifest below manages the..."

README.markdown
```diff
@@ -187,49 +233,76 @@ If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicit
...
-Provides the contents of the specified value. Valid options: a string by default; an array if specified through the `type` parameter.
+``` puppet
+registry_value { 'title\value':
+  path => 'HKLM\Software\foo',
+  ...
+}
+```
 
-#### `registry::service`
+The manifest below would manage the default value (extracted from the title because value_name is not specified) in the key 'HKLM\Software\foo'
```
clairecadman: Suggestion: omit would. "The manifest below manages the..."

README.markdown
```diff
@@ -187,49 +233,76 @@ If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicit
...
+*Optional.* Specifies a Registry value for Puppet to manage. Valid options: a string containing a Registry path. If any of the parent keys in the path don't exist, Puppet creates them automatically. For example: `HKLM\Software` or `HKEY_LOCAL_MACHINE\Software\Vendor`. Default value: the subkey in the title of the resource
 
-##### `command`
+If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicitly managed using a prefix. For example: `32:HKLM\Software\Vendor`.
 
-Specifies the command to execute when starting the service. Valid options: a string containing the absolute path to an executable file.
+##### `value_name`
 
-##### `start`
+*Optional.* Specifies the name of the registry value to manage.  For example: 'Value1' or 'Value\1'. This is typically specified separately when the value name includes backslashes. A value of empty string, for example '' denotes the default value of the registry key path will be managed. Default value: the value name in the title of the resource
```
clairecadman: Add comma after "for example".

Add "that" - "...for example, denotes that the default value..."

Add full stop after "resource."

README.markdown
```diff
@@ -187,49 +233,76 @@ If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicit
...
+All parameters are required unless otherwise stated.
 
-##### `description`
+##### `path`
 
-*Optional.* Provides a description of the service. Valid options: a string. Default value: blank.
+*Optional.* Specifies a Registry value for Puppet to manage. Valid options: a string containing a Registry path. If any of the parent keys in the path don't exist, Puppet creates them automatically. For example: `HKLM\Software` or `HKEY_LOCAL_MACHINE\Software\Vendor`. Default value: the subkey in the title of the resource
 
-##### `command`
+If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicitly managed using a prefix. For example: `32:HKLM\Software\Vendor`.
```
clairecadman: Suggested rewording (active voice over passive):

"If Puppet is running on a 64-bit system, manage the 32-bit Registry key using a prefix".

clairecadman:
@glennsarti added a few comments for suggested changes. But other than those it looks good to go.

README.markdown
glennsarti: Changed.  Also expanded on the note below that the path and value_name parameters will be used in preference to the resource title (implicit vs explicit settings)

README.markdown
glennsarti: Fixed

README.markdown
glennsarti: Fixed

README.markdown
glennsarti: Fixed

README.markdown
glennsarti: Fixed

README.markdown
glennsarti: Fixed

README.markdown
glennsarti: Fixed

README.markdown
glennsarti: Fixed

glennsarti:
@clairecadman Docs updated.

README.markdown
```diff
@@ -179,57 +225,84 @@ Tells Puppet whether the key should or shouldn't exist. Valid options: 'present'
...
 
-##### `type`
+``` puppet
+registry_value { 'HKLM\Software\Vendor\\':
+  ...
+}
+```
 
-*Optional.* Sets the data type of the specified value. Valid options: 'string', 'array', 'dword', 'qword', 'binary' and 'expand'. Default value: 'string'.
+Note you can explicitly set the path and value using the `path` or `value_name` parameters. These are used instead of the resource title.  When explicitly setting a `path` or `value_name` parameter, use both to avoid unexpected behavior, for example:
```
Iristyle: Shouldn't we just make it mandatory to specify both together?

README.markdown
```diff
@@ -179,57 +225,84 @@ Tells Puppet whether the key should or shouldn't exist. Valid options: 'present'
...
 #### `registry_value`
 
-##### `path`
+Manages individual Registry values.
 
-*Optional.* Specifies a Registry value for Puppet to manage. Valid options: a string containing a Registry path. If any of the parent keys in the path don't exist, Puppet creates them automatically. For example: `HKLM\Software` or `HKEY_LOCAL_MACHINE\Software\Vendor`. Default value: the title of your declared resource.
+The title of the resource determines the path and value name that will be managed, for example:
 
-If Puppet is running on a 64-bit system, the 32-bit Registry key can be explicitly managed using a prefix. For example: `32:HKLM\Software\Value3`.
+The manifest below manages the value called 'foo' in the key 'HKLM\Software'
```
Iristyle: Nit - I think you want `HKLM\Software` instead of 'HKLM\Software' to be consistent with rest of README

README.markdown
```diff
@@ -179,57 +225,84 @@ Tells Puppet whether the key should or shouldn't exist. Valid options: 'present'
...
 
-##### `ensure`
+``` puppet
+registry_value { 'HKLM\Software\foo':
+  ...
+}
+```
 
-Tells Puppet whether the value should or shouldn't exist. Valid options: 'present' and 'absent'. Default value: 'present'.
+You can also manage the default value of a key by adding a backslash at the end of the path.  The manifest below manages the default value in the key 'HKLM\Software\Vendor'
```
Iristyle: Nit - I think you want `HKLM\Software` instead of 'HKLM\Software' to be consistent with rest of README

README.markdown
```diff
@@ -179,57 +225,84 @@ Tells Puppet whether the key should or shouldn't exist. Valid options: 'present'
...
+registry_value { 'HKLM\Software\Vendor\\':
+  ...
+}
+```
 
-*Optional.* Sets the data type of the specified value. Valid options: 'string', 'array', 'dword', 'qword', 'binary' and 'expand'. Default value: 'string'.
+Note you can explicitly set the path and value using the `path` or `value_name` parameters. These are used instead of the resource title.  When explicitly setting a `path` or `value_name` parameter, use both to avoid unexpected behavior, for example:
 
-##### `data`
+The manifest below manages the value called 'value' (extracted from the title because `value_name` is not specified) in the key 'HKLM\Software\foo'
```
Iristyle: Nit - I think you want `HKLM\Software` instead of 'HKLM\Software' to be consistent with rest of README

README.markdown
```diff
@@ -179,57 +225,84 @@ Tells Puppet whether the key should or shouldn't exist. Valid options: 'present'
...
-Provides the contents of the specified value. Valid options: a string by default; an array if specified through the `type` parameter.
+``` puppet
+registry_value { 'title\value':
+  path => 'HKLM\Software\foo',
+  ...
+}
+```
 
-#### `registry::service`
+The manifest below manages the default value (extracted from the title because `value_name` is not specified) in the key 'HKLM\Software\foo'
```
Iristyle: Nit - I think you want `HKLM\Software` instead of 'HKLM\Software' to be consistent with rest of README

README.markdown
```diff
@@ -179,57 +225,84 @@ Tells Puppet whether the key should or shouldn't exist. Valid options: 'present'
...
+**Parameters**
 
-*Optional.* Provides a Display Name for the service. Valid options: a string. Default value: the title of your declared resource.
+All parameters are required unless otherwise stated.
 
-##### `description`
+##### `path`
 
-*Optional.* Provides a description of the service. Valid options: a string. Default value: blank.
+*Optional.* Specifies a Registry value for Puppet to manage. Valid options: a string containing a Registry path. If any of the parent keys in the path don't exist, Puppet creates them automatically. For example: `HKLM\Software` or `HKEY_LOCAL_MACHINE\Software\Vendor`. Default value: the subkey in the title of the resource
```
Iristyle: Note backtick wrappers here for `HKLM\Software`

README.markdown
glennsarti: AFAIK I can't enforce them because `value_name` allows empty values.

glennsarti:
Closed in favour of #146

PR #143, opened 2017-12-13
(MODULES-2957) Fix registry_value tests

glennsarti:
Previously the tests for the registry_value type were incomplete or were not
testing what they said they were.  This commit;
* Adds a compile test for when one resource is setting a default value and
  another is not, but shares the same name.  A fix for this was implemented
  in 64bba67 however there were no tests for this behaviour
* Fixed a bug in the canonicalising of path names which would append a backslash
  every time the method was called
* Actually tested that multiple backslashes are removed on the path
* Raised errors when an non default, but root only path was passed into a
  registry value.  This is different to registry keys, where this is allowed.
* Added tests for autorequiring of registry_key from registry_value resources
* Added tests in the registry_value provider for creating and destroying default
  values
* Added registry_value provider tests for case-insensitive exists? method and
  added for creating and destroying default values
* Fixed typo in the registry_key type tests whereby it was actually testing the
  resource aliases, not the autorequire of the type

### DIFF
--- a/spec/classes/mixed_default_settings_spec.rb
+++ b/spec/classes/mixed_default_settings_spec.rb
@@ -1,0 +1,25 @@
+require 'spec_helper'
+
+# The manifest used for testing, is located in spec/fixtures/mixed_default_settings/manifests/init.pp
+
+# This test is to ensure that the way of setting default values (trailing slash) compiles correctly
+# when there is a similar looking path in the manifest. It mixes default and non-default value_names in the same manifest
+#
+# This manifest attempts to manage two registry values
+#
+# + HKLM
+#     + Software
+#         - foo                  <-- This is a value called 'foo' in the 'HKLM\Sofware' key.
+#         |                          This is the 'hklm\Software\foo' resource
+#         |
+#         + foo
+#             + (default value)  <-- This is the default value for a key called 'HKLM\Software\foo'.
+#                                    This is the 'hklm\Software\foo\\' resource
+#
+
+describe 'mixed_default_settings' do
+  it { is_expected.to compile }
+
+  it { is_expected.to contain_registry_value('hklm\Software\foo\\') }
+  it { is_expected.to contain_registry_value('hklm\Software\foo') }
+end
--- a/spec/fixtures/mixed_default_settings/manifests/init.pp
+++ b/spec/fixtures/mixed_default_settings/manifests/init.pp
@@ -1,0 +1,13 @@
+class mixed_default_settings {
+  registry_value { 'hklm\Software\foo\\':
+    ensure => present,
+    type   => string,
+    data   => 'default',
+  }
+
+  registry_value { 'hklm\Software\foo':
+    ensure => present,
+    type   => string,
+    data   => 'nondefault',
+  }
+}
--- a/spec/unit/puppet/provider/registry_value_spec.rb
+++ b/spec/unit/puppet/provider/registry_value_spec.rb
@@ -51,7 +51,12 @@ describe Puppet::Type.type(:registry_value).provider(:registry), :if => Puppet.f
 
   describe "#exists?" do
     it "should return true for a well known hive" do
-      reg_value = type.new(:path => 'hklm\SOFTWARE\Microsoft\Windows NT\CurrentVersion\SoftwareType', :provider => described_class.name)
+      reg_value = type.new(:title => 'hklm\SOFTWARE\Microsoft\Windows NT\CurrentVersion\SoftwareType', :provider => described_class.name)
+      reg_value.provider.exists?.should be true
+    end
+
+    it "should return true for a well known hive with mixed case name" do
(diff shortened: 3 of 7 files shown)


glennsarti:
Running through adhoc

glennsarti:
Adhoc passed.  Ready for merge

spec/fixtures/mixed_default_settings/manifests/init.pp
```diff
@@ -0,0 +1,13 @@
+class mixed_default_settings {
```
Iristyle: I'm not sure what `mixed_default_settings` means in this context?

glennsarti: mixes default value_name and non-default valuename

`'hklm\Software\foo\\'` vs `'hklm\Software\foo'`

glennsarti: I'll add comments into the test to explain this.

glennsarti: Add an ASCII art reg tree to show what's going on

spec/unit/puppet/type/registry_value_spec.rb
```diff
@@ -60,12 +61,50 @@
...
+      }
+    ].each do |testcase|
+      context testcase[:context] do
+        let(:subject) { described_class.new(:title => testcase[:reg_value_title], :catalog => catalog) }
+
+        it 'should not autorequire ancestor keys if none exist' do
+          expect(subject.autorequire).to eq([])
+        end
+
+        it 'should only autorequire the immediate ancestor registry_key resource' do
```
glennsarti: This test is slightly wrong.  Autorequries grabs the first nearest parent key

spec/unit/puppet/type/registry_value_spec.rb
```diff
@@ -60,12 +61,50 @@
...
+          catalog.add_resource(Puppet::Type.type(:registry_key).new(:path => 'hklm\Software', :catalog => catalog))
+          catalog.add_resource(Puppet::Type.type(:registry_key).new(:path => 'hklm\Software\foo', :catalog => catalog))
+          catalog.add_resource(Puppet::Type.type(:registry_key).new(:path => 'hklm\Software\foo\bar', :catalog => catalog))
```
Iristyle: Add one new test here like:

```
catalog.add_resource(Puppet::Type.type(:registry_key).new(:path => 'hklm\Software\foo\bar\baz\ethan', :catalog => catalog))
```
And make sure that it autorequires nearest value of 'hklm\Software\foo\bar'

glennsarti:
@Iristyle Issues have been addressed.  I also pulled in some tests I created in #142 into this PR.  Updated the commit message accordingly.

glennsarti:
Ready for merge

spec/classes/mixed_default_settings_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+# This manifest attempts to manage two registry values
+#
+# + HKLM
+#     + Software
+#         - foo                  <-- This is a value called 'foo' in the 'HKLM\Sofware' key.
+#         |                          This is the 'hklm\Software\foo' resource
+#         |
+#         + foo
+#             + (default value)  <-- This is the default value for a key called 'HKLM\Software\foo'.
+#                                    This is the 'hklm\Software\foo\\' resource
```
Iristyle: Big improvement 👍

spec/unit/puppet/type/registry_value_spec.rb
```diff
@@ -60,12 +61,55 @@
     it 'should validate the length of the value data'
     it 'should be case-preserving'
     it 'should be case-insensitive'
```
Iristyle: We can worry about this later.. but any reason to leave these since they don't do anything either?

spec/unit/puppet/provider/registry_value_spec.rb
```diff
@@ -51,7 +51,12 @@
 
   describe "#exists?" do
     it "should return true for a well known hive" do
-      reg_value = type.new(:path => 'hklm\SOFTWARE\Microsoft\Windows NT\CurrentVersion\SoftwareType', :provider => described_class.name)
+      reg_value = type.new(:title => 'hklm\SOFTWARE\Microsoft\Windows NT\CurrentVersion\SoftwareType', :provider => described_class.name)
```
Iristyle: All the other tests in this file use `:path` rather than `:title` (given `:path` is the `namevar`) ... I assume this ties into the next PR? Just a little peculiar to see a different attribute for the type without any explanation for why `:title` is used here rather than `:path`.

Iristyle:
A couple minor comments we can look at later - good enough for now given other changes pending.

spec/unit/puppet/provider/registry_value_spec.rb
glennsarti: copy-pasta typo.  Should've been path

PR #146, opened 2017-12-27
(MODULES-2957) Add support for backslashes in registry value name

glennsarti:
Previously the registry module was compatible with older Puppet agent versions
(< 4.7.0) however as the module only now supports 4.7.0 and above, much of this
code can be removed.  In particular many of the FFI calls and lower level API
methods are now contained within Puppet itself.  These should be used instead of
the custom written ones in this module to reduce code duplication and reduce
the testing burden.  This commit removes duplicate code, mainly from the base
class with some in the providers.  This commit also moves the valuename method
back into the registry_value provider as it is not used in the registry_key
provider so there is no need for it to be shared.

---

(MODULES-2957) Add support for backslashes in registry value name

Previously it was not possible to modify a registry value with a backslash in
the name as the existing code used backslashes as the delimiter for registry
key paths.  As registry key names cannot have a backslash or be empty, the
delimiter between a registry key path and value is changed to a double backslash
for example:

HKLM\Software\Vendor\\ValueWithA\Backslash.

In this case the registry key path would be 'HKLM\Software\Vendor' and the value
name 'ValueWithA\Backslash'.

This commit:
* Uses a double backslash by default to determine the key path and valuename
  from a resource title, but will revert back to the old behaviour (single
  backslash) if none are found
* Adds tests to show the new functionality in the custom type (registry_value),
  the defined helper type (registry::value) and in the purging of registry_key
  resources

---

Previously when purging values, the search was case insensitive which meant
some registry values would be deleted or not be deleted in error.  This commit
changes the eval_generate method so it compares value lists in a case
insensitive way.


README.markdown
```diff
@@ -179,57 +224,72 @@ Tells Puppet whether the key should or shouldn't exist. Valid options: 'present'
...
-Tells Puppet whether the service should or shouldn't exist. Valid options: 'present' and 'absent'. Default value: 'present'.
+All parameters are required unless otherwise stated.
 
-##### `display_name`
+##### `path`
 
-*Optional.* Provides a Display Name for the service. Valid options: a string. Default value: the title of your declared resource.
+The path to the registry value to manage.  For example:
+'HKLM\Software\Value1', 'HKEY_LOCAL_MACHINE\Software\Vendor\Value2'. If Puppet is running on a 64-bit system, the 32-bit registry key can
+be explicitly manage using a prefix.  For example: '32:HKLM\Software\Value3'. Use a double backslash between the value name
```
clairecadman: Line break. Also managed*

README.markdown
```diff
@@ -179,57 +224,72 @@ Tells Puppet whether the key should or shouldn't exist. Valid options: 'present'
...
+      and path when managing a value with a backslash in the name.
```
clairecadman: Line break?

README.markdown
```diff
@@ -179,57 +224,72 @@ Tells Puppet whether the key should or shouldn't exist. Valid options: 'present'
...
 
-##### `data`
+``` puppet
+registry_value { 'HKLM\Software\Vendor\\\A\Backslash':
+  ...
+}
+```
 
-Provides the contents of the specified value. Valid options: a string by default; an array if specified through the `type` parameter.
+Note you must use '\\\' in the path because backslashes must be escaped if the next character is a backslash.
```
clairecadman: Sugested rewording:

Note that you must use '\\\' in the path. Backslashes must be escaped if the next character is a backslash.

README.markdown
glennsarti: Fixed. copy-pasta fail from the type documentation

README.markdown
glennsarti: Fixed. copy-pasta fail from the type documentation

README.markdown
glennsarti: Unfortunately that isn't the intent of my sentence.  I'll try a different re-wording.

glennsarti: Updated.

glennsarti:
@clairecadman  I added a new commit to the README to fixup conflicting information in the README.

.sync.yml
```diff
@@ -1,4 +1,7 @@
 ---
+.travis.yml:
```
Iristyle: Shouldn't need this anymore...

.travis.yml
```diff
@@ -6,7 +6,8 @@ cache: bundler
 script: "bundle exec rake release_checks"
 #Inserting below due to the following issue: https://github.com/travis-ci/travis-ci/issues/3531#issuecomment-88311203
 before_install:
-  - gem update bundler
+  # Disabled due to MODULES-6339
```
Iristyle: Shouldn't need here either

README.markdown
```diff
@@ -51,23 +50,30 @@ registry_value { 'HKLM\System\CurrentControlSet\Services\Puppet\Description':
...
   registry::value { 'puppetmaster':
-  key  => 'HKLM\Software\Vendor\PuppetLabs',
-  data => 'puppet.puppet.com',
+    key  => 'HKLM\Software\Vendor\PuppetLabs',
+    data => 'puppet.puppet.com',
   }
 }
 ```
 
 Puppet looks up the key 'HKLM\Software\Vendor\PuppetLabs' and makes sure it contains a value named 'puppetmaster' containing the string 'puppet.puppet.com'.
```
Iristyle: Should we backtick these (like the rest of the file / examples)?

README.markdown
```diff
@@ -51,23 +50,30 @@ registry_value { 'HKLM\System\CurrentControlSet\Services\Puppet\Description':
 }
 ```
 
+### Manage a single Registry value with a complex name
```
Iristyle: We jump right into the "complex name" example after the most basic example, though it's unclear what "complex" means here without a little more explanatory copy.

Should we start with a simpler example to bridge the gap?  For instance, instead of going straight to something with a `path` to the value as an attribute, could we start with no `path` attribute and an explanation of what it does?

```puppet
registry_value { 'HKLM\System\CurrentControlSet\Services\Puppet\\\ValueWithA\Backslash':
  ensure     => present,
  type       => string,
  data       => "The Puppet Agent service periodically manages your configuration",
}
```

Or something without the extra slashes that shows off `path`?

```puppet
registry_value { 'MyValue' :
  path         => 'HKLM\System\CurrentControlSet\Services\Puppet\Value'
  ensure     => present,
  type       => string,
  data       => "The Puppet Agent service periodically manages your configuration",
}
```

The main thing I'm getting it is that it's unclear from the examples when / why you need to separate things out.

README.markdown
glennsarti: single quotes for strings.  Backticks for lables/symbols

glennsarti:
Rebased onto master

README.markdown
glennsarti: What about

---

### Manage a single Registry value

``` puppet
registry_value { 'HKLM\System\CurrentControlSet\Services\Puppet\Description':
  ensure => present,
  type   => string,
  data   => "The Puppet Agent service periodically manages your configuration",
}
```

### Manage a single Registry value with a backslash in the value name

``` puppet
registry_value { 'HKLM\System\CurrentControlSet\Services\Puppet\\\ValueWithA\Backslash':
  ensure     => present,
  type       => string,
  data       => "The Puppet Agent service periodically manages your configuration",
}
```

### Manage a single Registry value with a different resource title

``` puppet
registry_value { 'PuppetDescription':
  path       => 'HKLM\System\CurrentControlSet\Services\Puppet\Description',
  ensure     => present,
  type       => string,
  data       => "The Puppet Agent service periodically manages your configuration",
}
```
```

Comment (Iristyle) [reply README.markdown]: Yeah, example  seems reasonable... think we need a little explanation between each? Thoughts @michaeltlombardi / @RandomNoun7 as "new users"?

Comment (michaeltlombardi) [reply README.markdown]: From a pseudo-beginner perspective, these all look fine. I'm able to pick up the difference between title and path from context even if I don't know about how Puppet works under-the-hood.

The other two are reasonably obvious examples, even for me. 😄

Comment (glennsarti) [reply README.markdown]: pushed

Comment (RandomNoun7) [reply README.markdown]: These examples seem good to me. Very clear.

**Comment by ThoughtCrhyme** [comment]
Has passed functional review here https://tickets.puppetlabs.com/browse/MODULES-2957?focusedCommentId=520384&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-520384

+1

File: lib/puppet/provider/registry_value/registry.rb [hunk 8]
```diff
@@ -24,8 +24,8 @@ def exists?
     found = false
     begin
       hive.open(subkey, Win32::Registry::KEY_READ | access) do |reg|
-        from_string_to_wide_string(valuename) do |valuename_ptr|
-          status = RegQueryValueExW(reg.hkey, valuename_ptr,
+        FFI::MemoryPointer.from_string_to_wide_string(valuename) do |valuename_ptr|
```
Iristyle: Use `FFI::Pointer` here instead of `FFI::MemoryPointer`.  The helper is defined on the base class.

lib/puppet/provider/registry_value/registry.rb
```diff
@@ -95,8 +86,8 @@ def regvalue
     unless @regvalue
       @regvalue = {}
       hive.open(subkey, Win32::Registry::KEY_READ | access) do |reg|
-        from_string_to_wide_string(valuename) do |valuename_ptr|
-          if RegQueryValueExW(reg.hkey, valuename_ptr,
+        FFI::MemoryPointer.from_string_to_wide_string(valuename) do |valuename_ptr|
```
Iristyle: Use `FFI::Pointer` here instead of `FFI::MemoryPointer`.  The helper is defined on the base class.

lib/puppet/provider/registry_value/registry.rb
```diff
@@ -201,7 +192,7 @@ def data_to_bytes(type, data)
   end
 
   def write(reg, name, type, data)
-    from_string_to_wide_string(valuename) do |name_ptr|
+    FFI::MemoryPointer.from_string_to_wide_string(valuename) do |name_ptr|
```
Iristyle: Use `FFI::Pointer` here instead of `FFI::MemoryPointer`.  The helper is defined on the base class.

lib/puppet/provider/registry_key/registry.rb
```diff
@@ -37,13 +37,7 @@ def destroy
     Puppet.debug("Destroying registry key #{self}")
```
Iristyle: When using `puppet resource` I'm getting an error like:


```
C:\source\puppetlabs-registry [pr/146 +7 ~0 -0 !]> bundle exec puppet resource registry_key 'HKLM\Software\Vendor\PuppetLabsTest_Value_Glenn2\SubKey1' --modulepath .\spec\fixtures\modules\
Error: Could not run: undefined local variable or method `hkeys' for Puppet::Type::Registry_key::ProviderRegistry:Class
C:\source\puppetlabs-registry [pr/146 +7 ~0 -0 !]> git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
C:\source\puppetlabs-registry [master...origin/master +7 ~0 -0 !]> bundle exec puppet resource registry_key 'HKLM\Software\Vendor\PuppetLabsTest_Value_Glenn2\SubKey1' --modulepath .\spec\fixtures\modules\
registry_key { 'HKLM\Software\Vendor\PuppetLabsTest_Value_Glenn2\SubKey1':
  ensure => 'present',
}
```

Iristyle:
We've also got some problems using leading or trailing backslashes in the name of the value.  Leading parses, but doesn't set values in the registry properly.  Trailing won't even parse properly:

```puppet
# there was note in PR somewhere that seemed to indicate this shouldn't work, but works fine
registry_value { "HKLM\\Software\\Vendor\\PuppetLabsTest_Value_Glenn2\\SubKey1\\\\ValueStringWithoutBackslash":
  data => "Should be a string phase=1",
}

# middle backslash works fine, performs as expected
registry_value { "HKLM\\Software\\Vendor\\PuppetLabsTest_Value_Glenn2\\SubKey1\\\\ValueString1\\Foo":
  data => "Should be a string phase=1",
}

# leading backslash in name of \ValueStringLeading is eaten
registry_value { "HKLM\\Software\\Vendor\\PuppetLabsTest_Value_Glenn2\\SubKey1\\\\\\ValueStringLeading":
  data => "Should be a string phase=1",
}

# trailing backslash in ValueString1Trailing\ completely fails during munging like:

#Error: Parameter path failed on Registry_value[HKLM\Software\Vendor\PuppetLabsTest_Value_Glenn2\SubKey1\\ValueString1Trailing\]: Munging failed for value "HKLM\\Software\\Vendor\\PuppetLabsTest_Value_Glenn2\\SubKey1\\\\ValueString1Trailing\\" in class path: Invalid registry key: HKLM\Software\Vendor\PuppetLabsTest_Value_Glenn2\SubKey1\\ValueString1Trailing

# registry_value { "HKLM\\Software\\Vendor\\PuppetLabsTest_Value_Glenn2\\SubKey1\\\\ValueString1Trailing\\":
#   data => "Should be a string phase=1",
# }
```

Iristyle:
Also, double backslashes in the name appear to confuse the munging as well:

```puppet
registry_value { "HKLM\\Software\\Vendor\\PuppetLabsTest_Value_Glenn2\\SubKey1\\\\ValueString1\\\\Foo\\\\Bar\\Hi\\Hey":
  data => "Should be a string phase=1",
}
```

lib/puppet/provider/registry_value/registry.rb
glennsarti: Done

lib/puppet/provider/registry_value/registry.rb
glennsarti: Done

lib/puppet/provider/registry_value/registry.rb
glennsarti: Done

lib/puppet/provider/registry_key/registry.rb
glennsarti: Fixed in the "refactor Provider Base as a Module" commit

glennsarti:
Fixed the middle, trailing and leading double slash issues and added tests as well.

glennsarti:
Running through adhoc seem as though Travis is blocked up.

glennsarti:
Adhoc is green!

ThoughtCrhyme:
New tests are glorious and new code passes FR. Added notes to JIRA ticket indicating as much. We're passing all previously failing manual tests now.

glennsarti:
@Iristyle 
img

spec/unit/puppet/provider/registry_value_spec.rb
```diff
@@ -88,15 +98,16 @@
...
+    let (:path) { path = "hklm\\#{puppet_key}\\#{subkey_name}\\#{valuename}" }
     def create_and_destroy(path, reg_type, data)
       reg_value = type.new(:path => path,
         :type => reg_type,
         :data => data,
         :provider => described_class.name)
       already_exists = reg_value.provider.exists?
       already_exists.should be_falsey
-
+#require 'pry'; binding.pry
```
Iristyle: Whoopsy daisy...

glennsarti: No idea what you're talking about :-)  Done.

lib/puppet_x/puppetlabs/registry.rb
```diff
@@ -6,6 +6,44 @@ module Registry
   # For 64-bit OS, use 32-bit view. Ignored on 32-bit OS
   KEY_WOW64_32KEY = 0x200 unless defined? KEY_WOW64_32KEY
 
+  # Helper methods used in the registry providers
+  # Guarded to only load on Windows due to Win32 constants
+  if Puppet.features.microsoft_windows?
```
Iristyle: I don't think I understand why the guard is needed given the constants wouldn't be used unless this code is actually called?

glennsarti: The spec tests load the provider code which loads this.

lib/puppet_x/puppetlabs/registry.rb
```diff
@@ -124,45 +151,79 @@ def filter_path
...
 
-    def subkey
-      if default?
-        filter_path[:trailing_path]
+
+    # Extract the valuename from the path and then munge the actual path
+    def initialize(path)
+      # Try finding the valuename via the double backslash method first
+      # and then revert to the old single backslash way
+      dbl_slash_idx = path.index('\\\\') || 0
```
Iristyle: I think all of the array index logic can be simplified by setting the limit param to 2 on `split`, and a destructuring assignment like this:

```
[1] pry(main)> path, valuename = "foo\\\\bar".split("\\\\", 2)
=> ["foo", "bar"]
[2] pry(main)> path
=> "foo"
[3] pry(main)> valuename
=> "bar"

[4] pry(main)> path, valuename = "foo\\\\bar\\\\".split("\\\\", 2)
=> ["foo", "bar\\\\"]
[5] pry(main)> path
=> "foo"
[6] pry(main)> valuename
=> "bar\\\\"

[7] pry(main)> path, valuename = "foo\\\\\\bar\\".split("\\\\", 2)
=> ["foo", "\\bar\\"]
[8] pry(main)> path
=> "foo"
[9] pry(main)> valuename
=> "\\bar\\"

[10] pry(main)> path, valuename = "foo\\\\\\bar\\baz\\\\".split("\\\\", 2)
=> ["foo", "\\bar\\baz\\\\"]
[11] pry(main)> path
=> "foo"
[12] pry(main)> valuename
=> "\\bar\\baz\\\\"

[13] pry(main)> path, valuename = "foo\\bar\\".split("\\\\", 2)
=> ["foo\\bar\\"]
[14] pry(main)> path
=> "foo\\bar\\"
[15] pry(main)> valuename
=> nil

[16] pry(main)> path, valuename = "foobar".split("\\\\", 2)
=> ["foobar"]
[17] pry(main)> path
=> "foobar"
[18] pry(main)> valuename
=> nil
```

With no double backslash, `valuename` is `nil`, in which case you can fall back to doing the same thing with the single backslash..

Also, `end_with?('\\')` is a bit easier to read than a case statement against `path[-1, 1]`

lib/puppet_x/puppetlabs/registry.rb
Iristyle: Hmm... will have to look closer.  Going to merge for now.

Iristyle:
I will put up a subsequent PR to clean up the `initialize` logic from my last comment.

Vetted this again locally, and all the manifests / command line puppet resource / etc testing that I tried worked properly.

Iristyle:
Updated PR is at #148

PR #149, opened 2018-01-22
(MAINT) Convert CHANGELOG to KAC standard

michaeltlombardi:
Prior to this commit the windows team was not tracking
changes prior to release in the changelog. Per the
KeepAChangelog standard, using an unreleased section
provides value to the users and developers both.
Using this section will make release prep easier and
allow users who want to leverage the module from master
to do so with clear knowledge of what changes have
been made to the code since the last release.

This commit adds the unreleased section to the CHANGELOG
and adds notes for the changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,267 +1,110 @@
-## 2017-03-06 - Supported Release 1.1.4
-### Summary
+# Changelog
+All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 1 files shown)


michaeltlombardi:
[Here](https://github.com/puppetlabs/puppetlabs-registry/compare/1.1.4...master)'s the comparison to master covered in this unreleased section.

PR #158, opened 2018-03-06
(MODULES-6744) Add testmode switcher

jpogran:


jpogran:
This has passed adhoc

spec/acceptance/should_tolerate_mixed_case_spec.rb
```diff
@@ -134,22 +134,25 @@
         /Registry_value\[HKLM.Software.Vendor.PuppetLabsTest_MixedCase_\w+\\SubKey1\\ValueString3\].ensure: created/,
         /Registry_value\[HKLM.Software.Vendor.PuppetLabsTest_MixedCase_\w+\\SubKey1\\ValueString4\].ensure: created/,
     ]
+
+    agent_is_x64 = false 
     windows_agents.each do |agent|
+      agent_is_x64 = is_x64(agent)
```
Iristyle: If the nodeset includes both 32-bit and 64-bit agents, this doesn't seem right... I think you'd want to filter the nodes for x64, then change the `execute_manifest` below to `execute_manifest_on` and only use 64-bit agents.

spec/acceptance/should_manage_values_spec.rb
```diff
@@ -198,116 +198,121 @@ def getManifest(keypath, vendor_path, phase)
   end
 
   context "Start testing should_manage_values" do
+    agent_is_x64 = false
+
     windows_agents.each do |agent|
+      agent_is_x64 = is_x64(agent)
```
Iristyle: With multiple nodes in the nodeset, last one wins, which doesn't seem accurate if there are both 64-bit and 32-bit nodes...

spec/acceptance/should_create_key_spec.rb
```diff
@@ -101,45 +101,48 @@
...
-      values_purged = values_purged_native + (is_x64(agent) ? values_purged_wow : [])
-
-      it 'Registry - Phase 1.a - Create some keys' do
-        apply_manifest_on(agent, phase1, get_apply_opts) do
-          keys_created.each do |key_re|
-            assert_match(key_re, @result.stdout, "Expected #{key_re.inspect} to match the output. (First Run)")
-          end
-          assert_no_match(/err:/, @result.stdout, 'Expected no error messages.')
+      agent_is_x64 = is_x64(agent)
+    end
```
Iristyle: Think this `end` can be removed, and the tests can iterate over each node in the nodeset as was done previously... now that `execute_manifest_on(agent,` can be used.

jpogran:
Replaced previous attempt with the new `execute_manfiest_on` methods

.sync.yml
```diff
@@ -1,6 +1,12 @@
 ---
 Rakefile:
   unmanaged: true
+Gemfile:
+  optional:
+    ':system_tests':
+      - gem: beaker-testmode_switcher
+        version: '<= 0.2.0'
```
Iristyle: This looks like an accident - I think we need the `0.4.0` release for `execute_manifest_on` here and in Gemfile

glennsarti:
@ThoughtCrhyme RE "smoke test for the module installing"  I thought we'd get that implicitly by testing on with master?

ThoughtCrhyme:
@glennsarti I don't think we will if we continue to guard on windows_agents like we do here: https://github.com/puppetlabs/puppetlabs-registry/pull/158/files#diff-5e01029f711c726dc52525d83b377506R104

glennsarti:
Ahh yeah...there is that 👎

jpogran:
Updated both gemfile and sync.yml with correct (newest) versions

.sync.yml
```diff
@@ -1,6 +1,12 @@
 ---
 Rakefile:
   unmanaged: true
+Gemfile:
+  optional:
+    ':system_tests':
+      - gem: beaker-testmode_switcher
+        version: '<= 0.4.0'
```
Iristyle: This sets a maximum version of `0.4.0`... but we want a minimum of `0.4.0`. Any of these would work:

* `~> 0.4` - pessimistically versioned to `< 1.0`
* `= 0.4.0` - pin
* `>= 0.4.0` - anything going forward

First option is probably the safe bet for now...

jpogran:
Updated with new version constraint

PR #159, opened 2018-03-11
(MODULES-6818) Fix types to no longer use unsupported proc title patterns

treydock:
This change will fix errors when running `puppet generate types`

Errors:
```
Error: /etc/puppetlabs/code/environments/production/modules/registry/lib/puppet/type/registry_key.rb: title patterns that use procs are not supported.
Error: /etc/puppetlabs/code/environments/production/modules/registry/lib/puppet/type/registry_value.rb: title patterns that use procs are not supported.
```


glennsarti:
@treydock Thanks for the PR and ticket!!

PR #161, opened 2018-05-18
(FM-6934) Update acceptance tests for master-agent scenario

glennsarti:
Previously the spec helper would only install the registry module on Windows
platforms however when testing master-agent scenarios, the module must also
exist on the non-Windows master.  This commit updates the spec helper to install
the module onto all hosts, not just Windows platforms.


glennsarti:
Running through adhoc.

glennsarti:
Passing adhoc.

PR #162, opened 2018-08-01
(MODULES-7534) Use case insensitive search when purging

glennsarti:
Previously when the `purge_values` parameter was set to true, it would fail on
registry values which contained an uppercase name, and already exsts.  This was
due to the array comparison using `.downcase` for the is_value elements, but
not doing the same for the should_values elements.

This commit changes the should_values elements to always be downcased thereby
making the later comparison valid.  This commit also updates the tests, which
would now fail if the fix was not applied.


Iristyle:
LGTM. I did a quick scan for other places where we might have missed a `downcase`, but didn't see any.

PR #173, opened 2019-01-18
(MODULES-5625) Fail on empty strings in REG_MULTI_SZ

glennsarti:
Previously it possible to pass in empty strings into REG_MULTI_SZ
registry types, however according to the Microsoft documentation [1]
this is not allowed.

This commit modifies the registry_value type to validate incoming
array infomration and to fail if this data is passed through.

This commit also adds tests for these scenarios.

[1] https://docs.microsoft.com/en-us/windows/desktop/sysinfo/registry-value-types


glennsarti:
Bouncing PR to re-trigger CI.  Travis was lost.

PR #180, opened 2019-06-18
MODULES-9195 - Handle corrupt registry_values in a graceful way

rjpearce:
Corrupt registry entries raise an exception when read this prevents which prevents the desired state being enforced.

Ideally there should be some kind of mechanism for these errors should be handled gracefully. 

If gracefully handling is not the default case then a new parameter could be added to allow the user to choose the desired behaviour when read errors are encountered.

The spec test is a WIP as rspec itself will raise an exception when values have invalid encoding. I will continue to work on this but comments, guidance and suggestions very welcome on how to do this.

### DIFF
--- a/lib/puppet/provider/registry_value/registry.rb
+++ b/lib/puppet/provider/registry_value/registry.rb
@@ -100,8 +100,14 @@ Puppet::Type.type(:registry_value).provide(:registry) do
           if Puppet::Util::Windows::Registry.RegQueryValueExW(reg.hkey, valuename_ptr,
             FFI::MemoryPointer::NULL, FFI::MemoryPointer::NULL,
             FFI::MemoryPointer::NULL, FFI::MemoryPointer::NULL) == 0
-            # Note - This actually calls read from Win32::Registry not Puppet::Util::Windows::Registry
-            @regvalue[:type], @regvalue[:data] = from_native(reg.read(valuename))
+            contents = ''
+            begin
+              # Note - This actually calls read from Win32::Registry not Puppet::Util::Windows::Registry
+              contents = reg.read(valuename)
+            rescue Encoding::InvalidByteSequenceError => e
+              Puppet.log_exception(e, "#{e} when reading registry value for '#{valuename}'")
+            end
+            @regvalue[:type], @regvalue[:data] = from_native(contents)
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/registry_value/registry.rb
```diff
@@ -100,8 +100,14 @@ def regvalue
           if Puppet::Util::Windows::Registry.RegQueryValueExW(reg.hkey, valuename_ptr,
             FFI::MemoryPointer::NULL, FFI::MemoryPointer::NULL,
             FFI::MemoryPointer::NULL, FFI::MemoryPointer::NULL) == 0
-            # Note - This actually calls read from Win32::Registry not Puppet::Util::Windows::Registry
-            @regvalue[:type], @regvalue[:data] = from_native(reg.read(valuename))
+            contents = ''
```
glennsarti: Not sure how I feel about the default being an empty string.  This means an empty string is now either a) an empty and valid byte sequence in the registry or b) any valid byte sequence of arbitrary length.

Currently can't be nil either as the `data` method above will make it think the value is absent.

TheRealEdwardCullen: Current behaviour is to error.

This PR will cause resource to indicate both Type and Data to be unset.

I had a stab at making changes [here](https://github.com/TheRealEdwardCullen/puppetlabs-registry/blob/master/lib/puppet/provider/registry_value/registry.rb), which will result in only the Data being shown as incorrect (unset).

If there's some way to more directly indicate that the value could not be read, that would be good. What about simply having the string result be set to "<< Error reading value >>" or similar? There must be **some** string value that would be so unlikely in the real world as to be improbable... What about adding a GUID - that should cause the string to be unique?

I also added additional debugging and changed the format of the error messages slightly (first time in Ruby/Puppet, so not sure how useful they are).

eimlav:
Hi Richard. Thanks for taking the time to submit a PR for this issue. 

Myself and a number of my colleagues have taken a look at the issue you are having. We ultimately decided that this is not something which we believe is suitable for the module. 

One of the main reasons for this is that it could mangle the workflows of other customers/end users who may need the exception to be thrown to notify them of the corrupted data. It could be argued that a parameter like you mentioned could be added allowing you to toggle the workaround but then this 
 has additional consequences:
- Idempotency is lost
- Reports are generated on every Puppet run showing a changed resource. The previous value of the resource will also be incorrect due to the data being corrupted and there being no way of knowing what it actually is
- Removes traceability of errors

Should you require this fix in the module, the best approach going forward would be to use a fork of the module instead which includes the fix. But ultimately the underlying issue is that the application that is corrupting the data in the first place should be fixed. Apologies that we cannot go forward with your change and best of luck in finding an alternative solution.

PR #216, opened 2020-07-29
(IAC-967) Append null terminators since it is no longer part of the data

sanfrancrisko:
Early ruby versions did not handle wide terminators correctly and would sometime
read past the end of the string causing segfaults. To prevent that, puppet
started embedding a wide terminator within the string. So the ruby string 'hi'
would be encoded as the three character string 'hi\0', and then that would be
encoded as UTF-16LE.

While this works, it creates ambiguity as to whether the "\0" character is part
of the data or not. For example, "hi\n\0".chop is "hi\n" whereas "hi\n".chop is
"hi", and is probably what was intended.

Ruby has since fixed its wide terminator handling, and as of Puppet 7, the
`wide_string` method no longer embeds a terminator in the string. So update the
registry module to explicitly add the terminator(s) before calling
RegSetValueExW`.

### DIFF
--- a/lib/puppet/provider/registry_value/registry.rb
+++ b/lib/puppet/provider/registry_value/registry.rb
@@ -178,16 +178,30 @@ Puppet::Type.type(:registry_value).provide(:registry) do
     raise error
   end
 
+  def wide_string_to_bytes(data)
+    bytes = Puppet::Util::Windows::String.wide_string(data).bytes.to_a
+    # versions prior to 7 embedded a wide null in the string content to work
+    # around ruby bugs, see PUP-3970
+    if Puppet::PUPPETVERSION[0].to_i >= 7
+      bytes << 0 << 0
+    else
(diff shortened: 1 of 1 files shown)


sanfrancrisko:
**Tests**
- [ ] CI tests passing
- [ ] Release checks with `Puppet 6` passing
- [ ] Release checks with `Puppet 7` passing
- [ ] Release checks on `puppetlabs-motd` module using `joshcooper/puppetlabs-registry@fd9ec3953be357316726465e5fb92d36c5634e23` with `Puppet 6` passing
- [ ] Release checks on `puppetlabs-motd` module using `joshcooper/puppetlabs-registry@fd9ec3953be357316726465e5fb92d36c5634e23` with `Puppet 6` passing

pmcmaw:
LGTM. Thanks @joshcooper :-)

PR #254, opened 2021-08-19
Add possibility to produce a detailed error message 

reidmv:
Previously, the validation procedures were occuring in the #munge
method. This was a problem because if #munge raises an exception, the
error displayed to the user is opaque. If the data input is not valid,
the error should be raised in the #validate method. This will ensure
that the output given to the user includes information about the class,
resource, and parameter which is in error.

### DIFF
--- a/lib/puppet/type/registry_value.rb
+++ b/lib/puppet/type/registry_value.rb
@@ -175,6 +175,15 @@ Puppet::Type.newtype(:registry_value) do
     end
   end
 
+  validate do
+    # To ensure consistent behavior, always require a value for the data
+    # property. This validation can be removed if we remove the default value
+    # for the data property, for all data types.
+    if property(:data).nil?
+      raise ArgumentError, "No value supplied for required property 'data'"
+    end
+  end
+
   # Autorequire the nearest ancestor registry_key found in the catalog.
   autorequire(:registry_key) do
     req = []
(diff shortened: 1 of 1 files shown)


binford2k:
@reidmv if I'm reading this correctly, it doesn't add or remove any functionality at all, it just moves it to become more visible. Is that correct?

reidmv:
@binford2k that is correct, or, that is the intention, yes. The objective is to eliminate the possibility of an error like the following from being encountered:

```
Error: Failed to apply catalog: The data must be a valid DWORD:
```

And instead return an error like this one:

```
Error: Validation of Registry_value[HKLM\System\Value] failed: No value supplied for required property 'data' (line: 1)
```

Note that I've just added another two commits. I discovered that the real problem wasn't where the error was raised, munge vs. validate, but that the default value for the property wasn't passing validation when type was set to dword or qword. So, this commit now does include an additional internal change: it removes the invalid default value for those data types, but adds a type-level validation to require that the user set the property.

Unfortunately, it doesn't look like there's currently an easier way to produce a sane error message without changing the behavior of registry_value in some way (even if minor). Puppet doesn't seem to have been built to accommodate the scenario of a type supplying a default value that doesn't pass its own validation routines... 🤷

daianamezdrea (approved):

michaeltlombardi (approved):

daianamezdrea:
Thank you for your contribution, @reidmv !

PR #283, opened 2023-04-28
(MODULES-11424) Fix for wrong path in require

Rewerson:
Fix for error like:
Puppet Evaluation Error: Error while evaluating a Resource Statement, Could not autoload puppet/type/registry_value: no such file to load -- /etc/puppetlabs/code/environments/production/modules/registry/lib/puppet/type../../puppet_x/puppetlabs/registry


LukasAud:
Sorry for the long delay in approval.

Clebam:
Hi, I'm facing the same issue and even though I couldn't pinpoint the issue exactly, I don't think the fix is the right one. (even though it works)

For instance all dsc ressources uses the same call and runs without issue. (for instance : https://github.com/puppetlabs/puppetlabs-dsc/blob/main/lib/puppet/type/dsc_adcscertificationauthority.rb#L4 )

I really can't explain why it works though because the only instances of the path ending with a '/' when you request 'dirname' is when you're at the root ('/' on linux, 'C:/' on windows for instance)

Moreover it seems that only windows related module uses this call to "Pathname.new"

I'm still trying to figure out how and why this work with other modules, but I am not sure the fix proposed is ok. (Besides the fact that it works.)

Clebam:
FYI I ask puppet community slack (https://puppetcommunity.slack.com/archives/C0W1X7ZAL/p1685634200481399) and was told to use `require_relative` 
https://github.com/puppetlabs/puppetlabs-iis/blob/d2765c2c13755e56ed440de1e44f043c0c88c9fe/lib/puppet/type/iis_application.rb#L4

(I still can't explain why the dsc resources work)

Clebam:
FYI I opened a PR on the same subject (I couldn't find a way to contribute here :( ) https://github.com/puppetlabs/puppetlabs-registry/pull/287

PR #287, opened 2023-06-06
(MODULES-11424) Fix requirement method

Clebam:
This is related to this PR : https://github.com/puppetlabs/puppetlabs-registry/pull/283

Having discussed this subject on puppet slack, it is recommended to use require_relative for this purpose.

# The problem
 require fails to find the path because of concatenation.

# The cause
This fix : https://github.com/puppetlabs/puppetlabs-registry/commit/2e11fe6798ecd8ea5d46ea133bb70cc71f2452da
The thing is "Pathname" override the behavior of the "+" sign to handle the missing "/" in the path ( https://ruby-doc.org/stdlib-2.6.3/libdoc/pathname/rdoc/Pathname.html#method-i-2B )

# The fix
As mentionned, it is recommended to use `require_relative`

### DIFF
--- a/lib/puppet/provider/registry_key/registry.rb
+++ b/lib/puppet/provider/registry_key/registry.rb
@@ -1,15 +1,7 @@
 # frozen_string_literal: true
 
-# REMIND: need to support recursive delete of subkeys & values
-begin
-  # We expect this to work once Puppet supports Rubygems in #7788
-  require 'puppet_x/puppetlabs/registry'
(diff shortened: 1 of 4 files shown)


jordanbreen28:
Thanks @Clebam!
Using `require_relative` is the preferred approach here, I've kicked the tests and lets 👀 what happens

jordanbreen28 (approved):

PR #316, opened 2025-09-18
Add support for Sensitive data in registry_value

buminks:
## Summary
In the case, that you want to store a credential in registry, the credential is shown in clear text in log / puppetboard.
This change adds the support for Sensitive data in the registry module. So credentials are now shown in the log anymore.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified. (For example `puppet apply`)

PR #319, opened 2026-04-16
(PA-8354): Add support for Sensitive data in registry_value

span786:
- Updated REFERENCE.md to include Sensitive[String] in the data type options.
- Enhanced registry_value provider to unwrap Sensitive values before processing.
- Modified registry_value type to handle Sensitive data during value munging.
- Updated value.pp documentation to demonstrate usage of Sensitive types.
- Added unit tests to verify support for Sensitive data across different types.

PR created to run acceptance test against PR #316

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -164,11 +164,13 @@ Data type:
 Optional[Variant[
       String,
       Numeric,
-      Array[String]
+      Array[String],
+      Sensitive[String]
   ]]
 ```
 
-The data to place inside the registry value.
+The data to place inside the registry value. Can be a String, Numeric, Array[String],
+or Sensitive[String] for sensitive data like passwords.
 
 Default value: `undef`
 
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -164,13 +164,14 @@ Data type:
 Optional[Variant[
       String,
       Numeric,
-      Array[String],
+      Array[Variant[String, Sensitive[String]]],
       Sensitive[String]
   ]]
 ```
 
-The data to place inside the registry value. Can be a String, Numeric, Array[String],
-or Sensitive[String] for sensitive data like passwords.
+The data to place inside the registry value. Can be a String, Numeric,
+Array[Variant[String, Sensitive[String]]], or Sensitive[String] for
+sensitive data like passwords.
 
 Default value: `undef`
 
@@ -260,4 +261,3 @@ The path to the registry value to manage.
 
 The specific backend to use for this `registry_value` resource. You will seldom need to specify this --- Puppet will
 usually discover the appropriate provider for your platform.
-
--- a/examples/sensitive_example.pp
+++ b/examples/sensitive_example.pp
@@ -1,0 +1,26 @@
+# Example demonstrating Sensitive[String] support in registry::value
+# This example shows how to use Sensitive types for sensitive data like passwords
+
+# Create a sensitive password value
+$sensitive_password = Sensitive('mysecretpassword123')
+
+# Use the sensitive password in a registry value
+registry::value { 'DefaultPassword':
+  key  => 'HKLM\Software\MyApp',
+  data => $sensitive_password,
+  type => 'string',
+}
+
+# You can also use it directly inline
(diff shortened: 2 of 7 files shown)


skyamgarp:
## Code review issues

- **Critical bug**: the `data` property uses `array_matching: :all`, whose `insync?` in Puppet core never calls `property_matches?` — it does a raw `is == @should` comparison instead. Since `Sensitive#==`/`#to_s` never match plaintext, any `Sensitive`-wrapped `data` value is **always reported out of sync**, so Puppet rewrites the registry value on every single run (never converges). Verified empirically: plain strings converge fine via `insync?`; `Sensitive`-wrapped values never do, even when the underlying value already matches.
- As a consequence, all the new sensitive-aware logic in `property_matches?` and `normalize_comparable` (`lib/puppet/type/registry_value.rb:125-130, 195-210`) is **dead code** — never executed by a real Puppet run.
- The new `property_matches?` unit test calls the method directly with full arrays, bypassing `insync?` entirely — gives false confidence and doesn't catch the bug above.
- No idempotency/"apply twice" test exists (unit or acceptance) that would have caught this.
- Edge case: a raw `registry_value` resource with `data => Sensitive(['a','b'])` (whole array wrapped, not per-element) triggers a misleading "can only contain string values" error in the `:array` validate branch.
- Provider's `to_native` unwrap (`lib/puppet/provider/registry_value/registry.rb`) is redundant with the type's own unwrap/rewrap — harmless but uncommented as such; also has zero dedicated test coverage (though no provider specs exist in this repo at all pre-PR).
- Minor style: redundant comment `# Unwrap Sensitive values if present` restates the code.

**Recommendation:** hold this PR until the `insync?`/idempotency bug is fixed and covered by a test that exercises `insync?` (or an acceptance test doing two applies), not just `property_matches?` directly.

---
🤖 Generated by [Claude Code](https://claude.com/claude-code)

span786:
Fixed in a7d5b5b.

You were right — `array_matching: :all` makes `Puppet::Property#insync?` do a raw `is == @should` comparison and skip `property_matches?` entirely, so any `Sensitive`-wrapped `data` was always reported out of sync and rewritten on every run.

Overrode `insync?` on the `data` property to zip `is`/`@should` and delegate to `property_matches?` per element, which is what makes the existing Sensitive-aware `property_matches?`/`normalize_comparable` logic reachable at runtime.

Also added tests that exercise `insync?` directly (not `property_matches?` in isolation, which you correctly flagged as giving false confidence) covering string/array convergence, a mismatch case, and idempotency across two applies. Verified these catch the regression: reverting the `insync?` override locally makes 3 of the 5 new tests fail as expected.

Dropped the redundant `# Unwrap Sensitive values if present` comment in the provider's `to_native` too.

Full suite (316 examples) passes and rubocop is clean.

span786:
## Can these commits be split apart?

Short answer: no, they're dependent, not independent chunks of work.

Buminks' 3 commits (`2f38e3b`, `9064f35`, `7804710`) introduce Sensitive-data support in `registry_value.rb` — the munge/validate blocks, `REFERENCE.md` docs, and the `value.pp` example. The 5 commits that follow aren't a separate feature; they rewrite and patch that exact code:

- `ab3e93b` ("Preserve sensitive registry data") replaces buminks' munge/validate blocks wholesale, introducing `sensitive_value?`, `unwrap_sensitive`, `rewrap_sensitive`, `normalize_comparable`, and `property_matches?`.
- `f40fc23`, `bbb72c2`, `d74056c` incrementally patch bugs in that same code.
- `a7d5b5b` (mine) fixes the `insync?` convergence bug in that same code, and only works because those helper methods already exist.

Given that dependency, we have two options:

- **(A)** Leave PR #319 as-is (all 8 commits, already fixed and pushed).
- **(B)** Set up a stacked PR: reset #319 back to the 3 buminks commits (force-push, which temporarily reintroduces the known convergence bug) and open a second PR, based on that branch, with the 5 commits squashed into one.

I'd rather not overcomplicate this — let's go with **(A)** and keep things as they are.

amitkarsale (approved):

